PR #42, opened 2024-03-20
Add About page

outofambit:
- also removes manual slugs in the en/ folder of the text-detail collection
- and fixes text-detail routing to account for that


stalgiag (approved):
perfect in every way!

PR #60, opened 2024-03-23
Fall back to English example code if not present in other locales

davepagurek:
Resolves https://github.com/bocoup/p5.js-website/issues/58

This falls back to the `/en` version of an example's code if one isn't present in another locale, so that automatic PRs from GitLocalize that only include a translated mdx will still build.


davepagurek:
I verified that #56 works with this change in place, but am now reverting the change in this PR since I think that example does not exist in the new set of examples in https://github.com/bocoup/p5.js-website/pull/59 is merged.

PR #63, opened 2024-03-25
Add library, update schema

davepagurek:
- Adds another library from the public repo
- Updates the categories to accommodate a rename we did in the public repo
- Updates the schema to allow multiple authors


davepagurek:
Never mind, just going to add this to the other PR!

PR #65, opened 2024-03-26
Correct image import for collections

stalgiag:
see title


stalgiag:
@davepagurek just wanted to flag you on this one since it required some changes to the submitted Library YAML files. I also had to use a still image for the p5.glitch image. This PR shows the correct way to import optimized local images in content frontmatter/YAML.

stalgiag:
Also, recognizing there is a lot to skim here - you can glean everything you need to know by looking at 'content/libraries/config.ts'. There was an update to the schema. Astro assets require a dedicated object for the static images. So now there is `featuredImage` and `featuredImageAlt` as separate properties

PR #66, opened 2024-03-26
Move example thumbnails, add link in frontmatter

davepagurek:
- Moves images into an `images` folder
- Adds `featuredImage` to fontmatter referencing the image
- Removes Spanish translation of example that is no longer present

@limzykenneth we still need to move assets to the public folder, but this should handle the thumbnails!


src/content/examples/config.ts
```diff
@@ -10,6 +10,7 @@ export const exampleSchema = z.object({
   // Aria label used for the live example code
   arialabel: z.string().optional(),
   relatedContent: relatedContent().optional(),
+  featuredImage: z.string(),
```
davepagurek, line 13: @stalgiag I was going to try to use the image format you were using instead of a string here, but I wasn't sure how to get that to work with the type inference when the schema is defined only inside `defineCollection`:
https://github.com/bocoup/p5.js-website/blob/13d7323b66596deea37b8eb25b0f0b35bb53eb83/src/content/types.ts#L20-L22

Do you know if there's a way to use that image type and get `z.infer` to work?

stalgiag: Oh interesting. I don't have the full context but what about something like this:

```ts
type SchemaFunctionArgs = {
  image: ImageFunction;
};

/**
 * Schema for the Example content type.
 */
export const exampleSchema = ({ image }: SchemaFunctionArgs) =>
  z.object({
    // Title of the example
    title: z.string(),
    // Aria label used for the live example code
    arialabel: z.string().optional(),
    relatedContent: relatedContent().optional(),
    featuredImage: image(),
  });

/**
 * Content collection for the Examples section of the site.
 * Each file represents a single example.
 */
export const examplesCollection = defineCollection({
  type: "content",
  schema: exampleSchema,
});
```

stalgiag: Oh nevermind, I see now that this makes the `z.infer` unhappy. Let me try something else

stalgiag: Okay, yeah that's a hairy one. There is an offramp in the form of directly typing instead of inferring the type. I would recommend doing that for now if it unblocks you. I can try to look for more elegant solutions tomorrow if you didn't find anything
```ts
type ImageType = {
  src: string;
  width: number;
  height: number;
  format: "png" | "jpg" | "jpeg" | "tiff" | "webp" | "gif" | "svg" | "avif";
};

interface ExampleData {
  title: string;
  arialabel?: string;
  featuredImage: ImageType;
}
// Define the type for content items that include data according to exampleSchema
export type ExampleDocContentItem = AstroBaseContentType<ExampleData>;
```

davepagurek: That gets the job done! I'll do that for now.

PR #67, opened 2024-03-26
Convert Contributors Conference 2019

davepagurek:
This converts an initial event page.
- I added a `YouTubeEmbed` component that we can style later
- I also added a `Logos` component that you can put a bunch of image links into and it applies styles to
- @stalgiag I updated the schema to add a "to" date in case the event specifies a from-to range, and a location

@Qianqianye let me know if this looks good!


Qianqianye:
Thanks @davepagurek!

PR #72, opened 2024-03-27
[p5.js KO] Create ko folder in reference

yinhwa:
For a better workflow for KO translators!


limzykenneth:
I'll merge the ones by GitLocalize actually since they do the same thing and don't create the empty file here.

yinhwa:
thanks so much @limzykenneth ! will close down this one!

PR #74, opened 2024-03-27
[p5.js KO] Create ADD.mdx for ko/p5 init

yinhwa:
Create ADD.mdx for ko/p5 init

P.S. As for KO translators, we'll be PRing with a sum of commits together per translator, instead of an individual commit like this! @limzykenneth @Qianqianye


yinhwa:
Closing as #70 being merged

PR #102, opened 2024-03-29
Update page name and one-line page description

Qianqianye:
We have communicated with the designers that we will remove the word 'Archive' in all the page names. The section name and its respective page name would be the same, for example 'Libraries Archive' would just be called 'Libraries'. Not sure if you still like to keep both entries in the UI list.

CC @stalgiag @outofambit


outofambit:
thanks @Qianqianye!

Issue #68, opened 2024-03-27
Reference file name clash


limzykenneth:
It is an edge case that needs to be deal with but essentially the translators noticed that when trying to translate `smooth()` the file it is pointing to is actually the [reference](https://github.com/bocoup/p5.js-website/blob/main/src/content/reference/en/p5/smooth.mdx) for the constant `SMOOTH` instead. The reference for `smooth()` is presumably overwritten.

The current website handles it with the capitalization difference but with file based routing here, is there some way to handle both `smooth()` and `SMOOTH` given some file system can't handle case sensitivity?

limzykenneth:
@stalgiag @outofambit

stalgiag:
Thanks for catching this! We can look at putting all of the global consts into a separate directory. I may not be able to get to this till next week. Would this be a major disruption to the localization process?

limzykenneth:
@stalgiag It is blocking some of those entries from being translated as we are translating functions and not constants at the moment. Not a big problem for now though as we still have other entries to translate anyway so next week is fine.

PR #110, opened 2024-03-30
Ref sections

davepagurek:
@stalgiag @outofambit For the reference section, we wanted to try out some slightly different structure than the existing p5.js site:
- rather than relying on an arbitrary ordering like the current site has, we have a specific order we'd like in order to show more generally relevant things first
- rather than hiding all class methods behind a class entry or search, we want to try having them visible on the reference page

I've added an array of modules in the order we want, and added a bit of code to split out the submodules + class methods. It's not pretty, but it has the category &rarr; subcategory/class &rarr; entry structure now.

image

### DIFF
--- a/src/content/reference/config.ts
+++ b/src/content/reference/config.ts
@@ -1,6 +1,28 @@
 import { z, defineCollection } from "astro:content";
 import { relatedContent } from "../shared";
 
+// Categories, ordered in a (rough) general-to-specific sequence for easier
+// reading. Some bits that we haven't finished revising are moved lower down
+// until we revisit them.
+export const categories = [
+  'Shape',
+  'Color',
+  'Typography',
+  'Image',
+  'Transform',
+  'Environment', // TODO: make new category for accessibility
+  '3D',
+  'Rendering',
+  'Math',
+  'IO',
+  'Events',
+  'DOM',
+  'Data',
+  'Structure', // TODO: move to top once revised
+  'Constants',
+  'Foundation',
+] as const
+
 const paramSchema = z.object({
   name: z.string(),
   description: z.string(),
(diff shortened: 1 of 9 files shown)


Qianqianye:
Thanks @davepagurek!

> * rather than hiding all class methods behind a class entry or search, we want to try having them visible on the reference page

To share some context on the second point, we are exercising the potential to see if it make sense to show all class methods  on the main reference page. We will need to see if it's too confusing to show entries like [pixelDensity()](https://p5js.org/reference/#/p5/pixelDensity) and [p5.Image/pixelDensity()](https://p5js.org/reference/#/p5.Image/pixelDensity) at the same page.
If it's too confusing and needs extra work, we are open to keep the current structure of hiding all class methods within a class entry for the upcoming website launch.

CC @limzykenneth too.

stalgiag:
> Thanks @davepagurek!
> 
> > * rather than hiding all class methods behind a class entry or search, we want to try having them visible on the reference page
> 
> To share some context on the second point, we are exercising the potential to see if it make sense to show all class methods on the main reference page. We will need to see if it's too confusing to show entries like [pixelDensity()](https://p5js.org/reference/#/p5/pixelDensity) and [p5.Image/pixelDensity()](https://p5js.org/reference/#/p5.Image/pixelDensity) at the same page. If it's too confusing and needs extra work, we are open to keep the current structure of hiding all class methods within a class entry for the upcoming website launch.
> 
> CC @limzykenneth too.

That makes sense. I am going to be working on styles, layout, big fixes, features etc related to Reference this week so your decision will be relevant to me. I'll leave off anything related to the class methods for now but it would be awesome if you could let me know what everyone is thinking by Thursday.

stalgiag (approved):
Thanks for this! Definitely needed! I am going to merge for now and use as a base for semi-related Reference work handling #68 and styles for individual Reference items. Other updates to the sorting/display on the Reference landing are welcome in the form of follow-ups.

PR #116, opened 2024-04-01
Prevent overwritten reference items

stalgiag:
This PR:

1. Stores parsed references for constants in a subfolder at 'p5/constants' to avoid overwriting methods with the same name
2. Removes now-obsolete constant entries in the p5 directory
3. Adds methods that were overwritten (ex: `smooth()`)
4. Ensures that the reference item paths do not include the new `constants` subdir
5. Removes constants without descriptions from the reference landing page
6. Minor reorganizing in related files

addresses #68 

cc @limzykenneth


stalgiag:
@outofambit I am going to go ahead and merge this since it is needed for the translation work but let me know if anything about it should be addressed in a follow-up

outofambit:
@stalgiag sounds good, apologies for the delay here. looks good to me, I'd like to document that normalize route helper, we can do that in a larger pass on docs for that helper file.

PR #117, opened 2024-04-01
Use OpenProcessing api for sketches pages

outofambit:
- Adds helpers to use the OpenProcessing API
- Uses API responses to populate the Sketches and individual Sketch pages
- Adds localization to those layouts
- Removes `src/content/sketches` since its no longer needed
- Adds a LinkButton and Icon component with basic styles
- Currently using a placeholder collection (with a limited size)

Will followup to do actual styling for these layouts in another PR


src/layouts/BaseLayout.astro
```diff
@@ -7,21 +7,23 @@ import ItemPageHeader from "@components/ItemPageHeader/index.astro";
 import { getCurrentLocale, useTranslations } from "@i18n/utils";
 import "@styles/base.scss";
 
-interface BaseLayoutProps {
+interface Props {
   title: string;
   className?: string;
-  variant: "root" | "item";
+  variant?: "root" | "item";
+  localize?: boolean;
```
outofambit, line 14: I don't love this solution but i think its useful for cases where we don't want/expect the title of the page to be localized. For example, the title of a sketch, or the name of a p5 method.

src/layouts/BaseLayout.astro
```diff
@@ -1,27 +1,31 @@
...
 import { getCurrentLocale, useTranslations } from "@i18n/utils";
 import "@styles/base.scss";
 
-interface BaseLayoutProps {
+interface Props {
   title: string;
   className?: string;
-  variant: "root" | "item";
+  variant?: "root" | "item";
+  localizeTitles?: boolean;
```
stalgiag, line 14: I guess I wonder if this is all to avoid [that console warning](https://github.com/bocoup/p5.js-website/blob/main/src/i18n/utils.ts#L147)? I almost feel like we should remove that anyway since it is so noisy during build and the fallback to the key is an important feature of the function.

src/pages/[locale]/sketches/[...slug].astro
```diff
@@ -1,39 +1,25 @@
...
 import { nonDefaultSupportedLocales } from "@i18n/const";
-import BaseLayout from "@layouts/BaseLayout.astro";
-import { getCollectionInLocaleWithFallbacks } from "@pages/_utils";
-import { removeLocalePrefix } from "@i18n/utils";
+import { getCurationSketches } from "@/src/api/OpenProcessing";
 
 export async function getStaticPaths() {
+  // this will be the same for all locales since these aren't localized
+  // TODO: remove limit once curation source is real
+  const sketches = await getCurationSketches(25);
```
Comment (stalgiag) [inline src/pages/[locale]/sketches/[...slug].astro:9]: Do we want to make that number 25 a const or the default value of the param? Just to make it easier for future editing to be sure that it is thorough?

src/layouts/CommunityLayout.astro
```diff
@@ -19,17 +26,15 @@ const t = await useTranslations(currentLocale);
...
   <h2>{t("Sketches")}</h2>
   {
-    sketches.map((entry) => (
+    sketches.slice(0, sketchLimit).map((sk) => (
       <li>
         <Image
-          src={entry.data.thumbnailImage}
-          alt={entry.data.thumbnailImageAlt}
+          src={makeThumbnailUrl(sk.visualID)}
+          alt="Screenshot of sketch"
```
stalgiag, line 33: Maybe a comment about this limitation on alt text?

stalgiag (approved):
Great! Small comments inline. Feel free to merge as you are ready.

outofambit:
thanks @stalgiag , i took all of suggestions!

PR #118, opened 2024-04-02
First pass on Reference item layout for method and constant items

stalgiag:
Note that these pages needs more work before it reaches "done" status but wanted to get this PR before the end of the day. I think the additional work should happen in a follow-up


outofambit (approved):
looks good! i am going to merge this now so i can work on top of some of these changes

PR #120, opened 2024-04-03
(Partially) Styled Homepage

outofambit:
~~(don't review + merge until after #117)~~


Qianqianye:
Hi @outofambit, we are planning the upcoming translation batch with website copy, and the text on the homepage would need to be updated.
We are wondering when you are planning to merge this PR. Thanks!

Qianqianye:
This timeline sounds good. Thanks @outofambit!

outofambit:
merging this even though styling isn't complete and will follow up with more soon. the structure of the page is set up and shouldn't change so translations should be able to happen @Qianqianye

PR #156, opened 2024-04-07
Specify order for tutorials within categories

davepagurek:
There's an order to the tutorials in some categories (it's optional overall, but some tutorials build off of previous ones.) I've added an optional `categoryIndex` to the config, set it on tutorials where the order matters, and have updated the layout to list them in that order.

Let me know if this method of specifying the order makes sense!


outofambit (approved):
this looks reasonable to me! thank you!

PR #158, opened 2024-04-07
Fetch extra library info from NPM

davepagurek:
One of the bits of info we collected from library authors was NPM info so that we could do things like automatically create a CDN script tag, fetch the license if not manually provided, and show users how actively developed the library is. This does some of that initial fetching, currently in a quick-and-dirty little list of tags:

image

@Qianqianye This additional info isn't currently in the designs because we were considering it a nice-to-have that we might not have had the ability to collect, but it looks like it's going to be feasible. Design-wise, what do you think the best approach for something like this should be? One option is to try to mimic the style of tags the designers had in some old iterations for things like examples, which we later took out from that section. Another option is to just hide this info for now and think of a more permanent placement at a later date.


davepagurek:
Update: the designers are OK with us using the tag design from [this file](https://www.figma.com/file/RZXPi1C3qvzNDWVGuBh5gj/STTY-%E2%80%94-Presentation-240212?type=design&node-id=15%3A3919&mode=design&t=cqNN19eeaNBSL4ii-1)
image

I've updated things slightly to look more in line with that, but also @stalgiag @outofambit we don't need this to hold up the rest of the site, so we're also OK just hiding this extra info for now if that makes things easier.

doradocodes:
@davepagurek Currently looking into merging this to `main`, but need to resolve merge conflicts. Since it's an old branch, I wanted to confirm what the expected the behavior of your change. From what I gather when looking at the current version of `LibrariesLayout.astro`, the current version works similarly to your change, but is using some newer components. Just want to make sure I don't override any of your work.

See latest version of `LibrariesLayout.astro` here: https://github.com/processing/p5.js-website/blob/main/src/layouts/LibrariesLayout.astro

davepagurek:
Hey! Generally the idea was to show some extra metadata for each library + have a link to quickly copy a script tag you can paste into your index.html. Some things that might need work before merging:
- We might want to revisit how these tags look in the libraries directory view (https://beta.p5js.org/libraries/directory/) which didn't exist back in 2024, and is generally more compact. Maybe a second row per library with these tags is ok?
- We probably need to check the current libraries again to make sure their npm info is actually correct / doesn't have typos / etc
- We may want to consider caching the npm data and checking it into the repo and periodically refreshing it? Less up-to-date that way, but there's the possibility that we get rate limited or something in the API changes and it starts blocking the build again simialr to what's been happening with OP. Maybe it's fine to leave as-is, but I could see that being a follow-up that could be needed.

PR #169, opened 2024-04-08
Community Page Styles

outofambit:
- **require a featured image for every past event**
- **better quality open processing images**
- **LinkButton supports passed classes**
- **community page mobile layout and styles**
- **hr styles**
- **desktop styles**
- **extract grid items into components**
- **add events layout**
- **rewrite sketches layout to use GridItemSketch**


src/layouts/CommunityLayout.astro
```diff
@@ -16,55 +15,79 @@ interface Props {
...
+      {
+        sketches.slice(0, sketchLimit).map((sk, i) => (
+          <li
+            class={(() => {
+              if (i === 0)
+                return "col-span-2 row-span-2 md:col-span-3 md:row-span-3";
+              if (i === 1) return "md:row-span-3";
+              if (i === 2) return "col-span-2 row-span-2";
+              if (i === librariesLimit - 1) return "row-span-2 col-span-2";
+              else return "md:row-span-2";
```
stalgiag, line 42: I don't know if this is a priority but his is a situation where using `nth-child` style selectors may be a little more friendly.

src/layouts/CommunityLayout.astro
```diff
@@ -16,55 +15,79 @@ interface Props {
...
+    <LinkButton url="/libraries" variant="link" class="mt-lg"
+      >{t("All Libraries")}</LinkButton
+    >
+  </section>
+  <hr />
+  <section class="mt-md mb-xl">
+    <h2 class="mb-lg">{t("Events")}</h2>
+    <ul class="grid grid-cols-2 gap-y-xl gap-x-md md:grid-cols-4">
+      {
+        pastEvents.slice(0, eventsLimit).map((event, i) => (
```
stalgiag, line 75: I know this hasn't been followed like a golden rule but it can make the markup easier to follow if any data transformation are done up in the "fences" and then used directly. Just so we can trust that the data that is rendered is only modified up there.

stalgiag (approved):
Looks good! Two small comments inline that may help with readability. I'll let you decide whether to address. We should also flag to the PF team that they will want to have some guarantees about thumbnail quality coming from OpenProcessing. Some of these images look very bad.

src/layouts/CommunityLayout.astro
outofambit: yeah fair point!

src/layouts/CommunityLayout.astro
outofambit: Yeah, I want to come back through and refactor these later. I think using an inline style tag in the layout would be good for stuff like this.

PR #170, opened 2024-04-08
Contribute Page styles

outofambit:
~~review and merge after #169~~

- adds styles for the Contribute page and styles for the individual Contributor docs pages
- includes a general "rendered markdown" class that takes care of spacing and some other basic styles that we can use for other places we just render a markdown doc
- updates contribute script to parse out a title and "description" out of every contributor doc
- had to update the contributor docs collection since i changed the frontmatter schema
- re-adds necessary language codes for contributor doc script to work

things to follow up on:
- check that parsing the first comment that usually explains what a contributor doc is is a good strategy (right now they aren't translated)
- some minor text overflow issues on the contribute page (the titles sometimes don't fit the box great at some sizes)
- use "rendered markdown" css class in more places


src/content/ui/en.yaml
```diff
@@ -46,3 +46,12 @@ briefPageDescriptions:
...
   People: Get to know the p5.js team.
+ContributePage:
+  statement: >
+    Our community values contributions of all forms and seeks to expand the meaning of the word contributor as far and wide as possible. It includes documentation, teaching, writing code, making art, writing, design, activism, organizing, curating, or anything else you might imagine.
+  imageAltText: >
+    A group of about 20 people standing and sitting together in front of a doorway with their hands up and happy looks on their faces
+  donateDescription: >
+    TEXT NEEDED HERE
+  faqDescription: >
+    TEXT NEEDED HERE
```
outofambit, line 57: @stalgiag do you think it would be better to have these in their own collection for just this page?

stalgiag: I think that could make sense. I would love a more consistent key naming strategy. Currently I believe the keys that lead to a value are pascal case and the keys that hold arrays are camel case.

styles/markdown.scss
```diff
@@ -0,0 +1,39 @@
...
+  h1 {
+    margin-top: var(--spacing-md);
+  }
+  ul {
+    list-style: disc;
+    padding-left: 2rem;
+    margin: var(--spacing-sm) 0;
+  }
+  ol {
+    list-style: georgian;
```
stalgiag, line 17: This looks cool but I think they just use decimal haha

Screenshot 2024-04-09 at 9 29 32 AM

styles/markdown.scss
```diff
@@ -0,0 +1,39 @@
+// Default styles to apply to markdown generated content
+
+@import "variables.scss";
+
+.rendered-markdown {
+  margin-bottom: var(--spacing-xl);
+
+  h1 {
+    margin-top: var(--spacing-md);
+  }
```
stalgiag, line 10: I find that it looks too balanced for most headings. As in the margin on top is equal to the margin below. It should be a larger space above, smaller below. It is subtle but it improves readability. First screenshot is this PR, second is Figma for reference

Screenshot 2024-04-09 at 9 30 34 AM
Screenshot 2024-04-09 at 9 38 12 AM

stalgiag (changes requested):
Looks good! Two small style comments and then I am good w approving

styles/markdown.scss
outofambit: loool 🤦🏼‍♀️

stalgiag (approved):

src/content/ui/en.yaml
outofambit: i'll standardize the naming for now and we can revisit if these should be in a different place once we get through a few more pages

PR #171, opened 2024-04-08
Image component with visible alt text and type fixes

stalgiag:
This PR:

1. Fixes the font family imports
2. Adds an image component wrapper that respects the "Show Alt Text" accessibility setting

Note that this will require switching all instances of `Image` imported from`astro:assets` to instead import from `@components/Image/index.astro`. I am waiting till some of the pending content changes land before doing a universal change in a follow-up.


src/components/Image/index.astro
```diff
@@ -0,0 +1,14 @@
+---
+import { Image } from "astro:assets";
+const { src, alt, width, height } = Astro.props;
```
outofambit, line 3: are the width and height considered optional? Astro's `Img` class should be able to read height and width of the image asset at build time so i don't think we should have to pass these in. it looks like you're using it like that, just want to be sure.

outofambit (approved):
looks good! we could also pull https://github.com/lyonsyonii/astro-preload into this at some point. left one question about props for the new component. as long as we're on the same page there then merge away :)

src/components/Image/index.astro
stalgiag: Yep I added some props to be more explicit

PR #175, opened 2024-04-09
Add VimeoEmbed component

davepagurek:
@Qianqianye usage:

```astro
import VimeoEmbed from '../../../components/VimeoEmbed/index.astro';

<VimeoEmbed id="169599296" />
```


Qianqianye:
Thanks @davepagurek!

PR #176, opened 2024-04-09
redo homepage as layout

outofambit:
Moved the homepage markup into its own route and created a one item collection for translations to live in.
Ported the existing translations to the new location.

We are really hitting the limits of images provided by the OpenProcessing here, and you'll see the grid breaks in one or two places because the images are just too small.
We will also need to get a version of the spinning asterisk gif that has less empty space baked into the image itself.


outofambit:
@Qianqianye @limzykenneth I needed to redo the homepage structure because it would have been a pain to maintain in the future. I moved the translations into their own folder in `src/content/homepage` and was sure to move the translations that have already been committed. Letting you know so you can redirect future translations to that folder. Happy to jump in on a PR if there are questions!

limzykenneth:
If possible can we merge this tomorrow after the content freeze just to we don't get potential merge conflict with translators currently working with the previous structure.

stalgiag (commented):
Looks great! 

The only issue that I noticed is that the spinning asterisk is too small in size on mobile. And the GIF file is WAAAAAAY too big even on fast internet it is a multiple second download. You should use [this SVG file](https://p5js.org/assets/img/thick-asterisk-alone.svg) and just animate the spin. 

Also probably a good idea to flag the issues with the OpenProcessing API images now. They are looking terrible

outofambit:
@limzykenneth yep, i can do that and handle any merge conflicts. no problem.

outofambit:
@stalgiag took care of that! @limzykenneth just confirming that i rebased and moved translation strings as needed.

PR #180, opened 2024-04-10
Nav, settings, theme updates, sidebar, jumpto links

stalgiag:
This PR adds the styles and interactions for the nav and theme settings menu. These are very complex UI elements with nuanced interaction and varied content so I can't say for certain that these are 100% done but they are in a good spot for now.

Since jump-to links are so varied in content, there is a global state for SSR called jumpToState that is set in each layout file. These are not set for every single layout yet and you may reach routes with stale content. I am waiting until more of the pages are in a "finished" state before a final universal "jump-to" pass


outofambit (commented):
looks good overall! this is a ton of work and i noticed some nice UI animations.

i found two issues i think we need to fix:
- i can't scroll the jump to list when its longer than the window
- we need to set `jumpToState(null)` on the text detail pages and a couple new ones that were added after this branch branched off of main. (they retain the jump to options of the previous page) could we have the `BaseLayout` default to `jumpToState(null)` if the wrapping template doesn't set it? or maybe the object should be passed to BaseLayout which then calls `jumpToState(null)` if something isn't passed.

i'm fine to have these addressed in this PR or in a followup right after

stalgiag:
> looks good overall! this is a ton of work and i noticed some nice UI animations.
> 
> i found two issues i think we need to fix:
> 
> * i can't scroll the jump to list when its longer than the window
> * we need to set `jumpToState(null)` on the text detail pages and a couple new ones that were added after this branch branched off of main. (they retain the jump to options of the previous page) could we have the `BaseLayout` default to `jumpToState(null)` if the wrapping template doesn't set it? or maybe the object should be passed to BaseLayout which then calls `jumpToState(null)` if something isn't passed.
> 
> i'm fine to have these addressed in this PR or in a followup right after

Thanks!

Former dealt with in the latest commit.

The latter should be addressed in a follow-up. I didn't want to finish all of the jumpState stuff while the other pages are in motion. And no, unfortunately, can't set jumpState(null) on the base layout because of DOM to fence execution order, the outermost element gets its fences processed last.

PR #209, opened 2024-04-16
Prevent flash of light theme on navigate, monochrome wordmark/logomark

### LINKED ISSUE #207: In dark mode, pages briefly flash to white when navigating
flash-white

I wonder if there's a way we can get a black background earlier in the loading process to avoid that?

stalgiag:
This PR prevents the flash of light theme on page navigate regardless of the applied theme. It does this by theming at the level of the `documentElement` instead and running a script before the body is mounted.

Unrelated to the primary aim of this PR but it also ensures that the logo variable updates for monochrome and dark mode.

fixes #207 

no-flash-p5


davepagurek (approved):
Nice!

PR #211, opened 2024-04-16
Fix search default locale links, locale switching on search results page, keep query params on updated search

stalgiag:
This PR does the following:

- Fix search links
- Keep query params when searching locales so that switching locales on the '/search' works
- Update query params when the search term is updated on the search page so that the router history is correct
- Remove reference consts without descriptions from search results


outofambit (approved):
looks good! tried this out locally

PR #212, opened 2024-04-22
Service worker caching, skeletal "offline mode"

stalgiag:
This PR adds a caching library and workbox config. It constrains HTML loading to prevent the downloading of thousands of HTML files upon first visit. Future iterations on this work could explore manual handling of the HTML caching to allow for download of the full site offline with constraints upon resource usage. Right now this defaults to caching only visited pages and limits the cache to 50 html files.

This will cache any assets loaded into an iframe on Firefox and Safari but unfortunately, chromium does not allow caching [via service worker or browser of any assets loaded into an iframe using the srcdoc attribute](https://github.com/w3c/ServiceWorker/issues/765). Further work will be needed to employ a different strategy for the code frame.

In the interim, this PR speeds up site performance significantly.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -3,6 +3,7 @@ import preact from "@astrojs/preact";
 import mdx from "@astrojs/mdx";
 import compress from "astro-compress";
 import tailwind from "@astrojs/tailwind";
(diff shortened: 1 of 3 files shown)


outofambit (approved):
incredible

PR #216, opened 2024-04-22
use real openprocessing curation collection

outofambit:


outofambit:
merging this because code changes are so minimal

PR #217, opened 2024-04-23
Header Components with Custom Topics

outofambit:
- makes "generated" topics work for localized pages
- adds ability to specify a topic from a layout for use in the header
- fixes spacing issues in headers


stalgiag:
Just pulling out that QA mentioned this issue in several different locations:

> In the header: Subtitle should be 24px and it is 30 px

Might be easiest to handle that with this work

stalgiag (approved):
Looks great! I like the breadcrumb solution, I'll do a pass using that on the other routes

PR #222, opened 2024-04-23
People page styling, placeholder content from mockups

stalgiag:
see title


outofambit (approved):
looks good! not sure my image needs to be 3MB but if that's what they gave us 🤷🏼‍♀️

PR #225, opened 2024-04-23
Download page

stalgiag:
Screenshot 2024-04-23 at 12 28 18 PM


outofambit (commented):
download page looks good! i wonder if we could use the `rendered-markdown` css class for the textdetail layout wrapper. the only issue i see is that the css changes break the linkbutton hover behavior. we might need a more scoped way to apply the a tag coloring or restructure the link button styles. i'm fine if we figure that out in a followup.

stalgiag:
Oh great point! I forgot that you made the `rendered-markdown` class. I went ahead and updated it to use that with a few tweaks for button link hover states.

PR #226, opened 2024-04-23
create RelatedItems component

outofambit:
- creates RelatedItems component
- uses in all places (except for the Event page since that will be styled soon and can be incorporated then)
- formalized a helper to get related items from a collection, utilizing our locale fallback logic


stalgiag (changes requested):
Looks great! The only thing that is not working is that the links themselves (at least for reference and examples) appear to be just the slug. I'd love to live in the world where that works but I think they will need the same transformation that is used at those respective path builders

outofambit:
@stalgiag thank you for catching that! should be good now!

stalgiag (approved):
Great! Thanks! I think we still need to handle the 'related' section in the reference items but that might require a slightly diff approach. Not sure but can happen in a follow-up and I can do it if you'd like

Issue #184, opened 2024-04-10
Translation of categories in Examples


stalgiag:
Some background, the "Examples" were originally created with the same folder structure as the examples for the current website. Categories are then derived from the folder name. As I was working on the jump to links for examples, I realized that this directory-derived category name is not translatable. 

A few possible solutions

1. Add the example category names to the UI translation doc
- Most direct
- Not ideal to put more data (especially data specific to a content collection) in the UI translation

2. Add a category property to the "Example" config
- Better organization
- Requires a single-line edit in each translation for all examples. 

All of this can happen after the content freeze but curious about preferences:

@Qianqianye @davepagurek @limzykenneth

PR #227, opened 2024-04-23
Add library categories and tag featured ones for the Community page

davepagurek:
- Updates the libraries readme in preparation for users submitting to this project in the future
- Links to the libraries readme from the main readme
- Ensures there is punctuation between the end of a library description and the "By {authors}" text
- Displays libraries by category on the libraries page, mostly copying the tutorials page code for that
- Adds a `featured` tag to libraries we want to show on the Community page
- Adds a few minor library updates from the other libraries repo

On the Community page:
image

On the All Libraries page:
image


src/content/libraries/config.ts
```diff
@@ -22,6 +22,27 @@ const categories = [
   "utils",
 ] as const;
 
+export const categoryNames: { [key in (typeof categories)[number]]: string } = {
```
stalgiag, line 25: Is there a reason why these can't be used in their directly renderable form in the frontmatter? I ask because this adds another place for translations when the translations could happen in the MDX if these were stored directly. see #230 for a related question about tutorials

davepagurek: I think the main thing we get by using slugs in each entry and then having translations of those is that we get easier type checking, since the options are the same for every locale folder, and then there's one less field that actually needs to be translated when new entries are made (the categories would just need to be translated once upfront.) So I've been thinking of them so far as something closer to UI translations than content translations.

That said, you're right that it's a new thing that needs to be translated. We can definitely still make it work with this being written ready to render in each file. We might need to do something sorta custom to make sure there aren't typos or discrepancies in the categories via Zod though -- is there a good way to do that without having a list of all the translations also stored somewhere so we can enforce those as options? I guess having that list centrally defeats the purpose of translating it in each file?

davepagurek: If you think the UI content approach makes sense, I can transition this + the tutorial category names into that file

stalgiag: That makes sense to me. I agree that it would be ideal to keep these as enum-like slugs. My concern was primarily about getting too spread out with translations. I think that moving them to a central location should suffice to make it clear that these utilities also need translations. Perhaps we can even do something similar for #184

davepagurek: That works for a way forward for #184! I just added some commits moving the library and tutorial category display names to the UI file.

stalgiag (approved):
This is fantastic. Thank you!!

PR #228, opened 2024-04-23
remove unused helper fn

outofambit:
noticed this while debugging something else


stalgiag (approved):
Makes sense to me. Feel free to merge once the build succeeds

PR #232, opened 2024-04-24
About Page and Text detail styling

outofambit:
- adds About page layout and styles
- removed extra padding in header on mobile (24 -> 20px)
- updates rendered markdown styles
- adds missing dark magenta text color
- the sum of this work and #229 should cover the text detail page styling


stalgiag (approved):
Nice!

PR #233, opened 2024-04-24
Add Related References for Reference Item Page

outofambit:
followup to #226


stalgiag (approved):
Love the backfilling from jump to links idea

PR #241, opened 2024-04-24
Event page styles

outofambit:
- removed placeholder event entries


stalgiag (commented):
Think this mostly gets it. Only thing that I noticed was off was that I think there is a `max-w-screen-lg` or `max-w-screen-md` on the content

PR #243, opened 2024-04-24
Fix code styles on Firefox Mobile

davepagurek:
Quick fix for the code styles on Firefox Mobile because I figure no one else is using this browser so I might as well debug it myself 😅 

Turns out it just does not have Courier or Courier New for some reason? But the fallback font is just `serif` instead of a monospaced font, so I've changed it here. I guess the variable name `--font-serif` is a bit of a misnomer, but I haven't updated that, just its value.

Before: 

After:


stalgiag:
Great! Thank you!

Issue #246, opened 2024-04-24
Contributor docs page entries


limzykenneth:
For the contributor docs, currently there are too many docs on the page as only content in the immediate folder should be used, the contents in `archive` and `project_wrapups` folders should not be used.

There are also a few files that should not be included such as `navbar.mdx` and `sidebar.mdx`. These are copied over from the main repo so I can delete them there to make sure they don't show up here, I can at the same time also delete the files here if you like.

PR #250, opened 2024-04-24
ignore some folders when importing contributor docs

outofambit:
see #244 and #246

~~enumerating these languages separately in the script is necessary to keep it from throwing these other languages into the `en` folder~~

adding a list of ignored folders to the contribute script to account for languages the website doesn't support and other folders we don't want to be included.

cc @stalgiag @limzykenneth


limzykenneth:
Perhaps related to #246, we don't really need anything within the subfolders, can we just grab the top level files and the supported languages only?

outofambit:
@limzykenneth oh thanks for pointing that out. i will revise this pr to handle that as well

stalgiag (approved):
Makes sense. Thank you!

PR #256, opened 2024-04-24
Fix for localized categories in reference

### LINKED ISSUE #245: Reference localization
I'm not sure if this is in the process of being address but currently the localized version of the reference is not looking quite right with many entries missing, from a glance, looks like all the translated ones? Notable excepting `rectMode()`.

Speaking of `rectMode()`, there's another issue that shows up in localization which is the one line description for Chinese.
Screenshot 2024-04-24 21:01:55

Because Chinese uses a different character for a full stop (`。`), simply truncating the first sentence before the first regular full stop character will cause the entire description to be rendered.

stalgiag:
Wrote a node script (not included since it was a one-time use) to extract the translated modules and submodules to fix #245

Now it looks like all of the translated reference links are included. I also handled the parsing of a single sentence using the fullstop characters for hindi and zh-Hans.

fixes #245

### DIFF
--- a/src/components/ReferenceDirectoryWithFilter/index.tsx
+++ b/src/components/ReferenceDirectoryWithFilter/index.tsx
@@ -1,7 +1,6 @@
 import type { ReferenceDocContentItem } from "@/src/content/types";
 import { useMemo, useState } from "preact/hooks";
 import type { JSX } from "preact";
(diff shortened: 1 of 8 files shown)


stalgiag:
Merging since this is important for other work but feel free to reopen the issue if there are additional problems

PR #270, opened 2024-04-25
Example assets path updates and fallback file processing

### LINKED ISSUE #267: Examples that load assets not working
[Hoisting this out of the Examples QA](https://github.com/processing/p5.js-website/issues/198#issuecomment-2059211345)

> Examples that load assets are not loading, e.g. https://p5js-wip.online/examples/imported-media-alpha-mask/

This is due to a combination of asset loading URLs not getting prefixed with "/" and some of the assets not being present in the public folder.

stalgiag:
fixes #267 

This does the following:
- moves all of the new example assets into the public folder
- adds a fallback to the example code loading function that ensures that all asset paths are prefixed with `/`
- modifies all example code assets to make them use the new pattern, ie: prefix asset path with `/`


outofambit (approved):
makes sense!

raccoon giving a thumbs up

PR #275, opened 2024-04-25
Meta attributes, performance improvements

stalgiag:
- Adds base meta attributes for all layouts
- Add `font-display: swap`
- Define `width` and `widths` in key locations to reduce image transfer size


stalgiag:
Merging since this work is needed for a11y work but will address any issues in a follow-up

Issue #221, opened 2024-04-23
Styling issues Tutorial 


ccanash:
DESKTOP 1280x800

1. Tutorials main page: https://p5js-wip.online/tutorials/
- [x] Add padding above and below in the header section - applies for all resolutions
- [x] In the images preview for all sub sections: reduce the spacing between the image title and subtitle. For example: between _Get Started_ and _A tutorial that [..]_
- [x] Remove grey line at the bottom of the page above the footer. Applies for all resolutions 

2. Tutorials sub pages: https://p5js-wip.online/tutorials/get-started/
- [x] The language, accessibility and search buttons are pink and should be brownish/gray like the color in the header of the tutorial main page 
- [x] At the bottom, related examples should have a line above and below, same for related references. 
- [x] The image that is a preview on the tutorials main page for each tutorial should appear at the top of each sub page. Applies for all resolutions. IT doesn't happen for get-started for example. 

TABLET 768x800
1. Tutorials main page: https://p5js-wip.online/tutorials/
- [x] In the header: Subtitle should be 24px and it is 30 px

2. Tutorials sub pages: https://p5js-wip.online/tutorials/get-started/
- [x] In the color-gradients tutorial the example code overflows the grey section to the right: https://p5js-wip.online/tutorials/color-gradients/ . 
![Screenshot 2024-04-23 at 10 50 21](https://github.com/bocoup/p5.js-website/assets/36164089/edbce5e4-0162-4fdb-865b-a1ed38f12c44) . This happens also for mobile resolution. 

MOBILE 428x800
1. Tutorials main page: https://p5js-wip.online/tutorials/
- [x] the description below each image should be 10px (but might be too small? ) It is currently 12px.

stalgiag:
Thank you!

>  The image that is a preview on the tutorials main page for each tutorial should appear at the top of each sub page. Applies for all resolutions. IT doesn't happen for get-started for example.

Not sure if this is the case since those images are different. I think that it makes more sense to leave this decision to the authors of the blog content but pulled it out to highlight it.

davepagurek:
Also, just noticed that the code examples no longer have JS syntax highlighting. @Qianqianye I see that in the Figma, tutorials don't have highlighting while examples do. Do you think it makes sense to just have highlighting everywhere? The code snippets in some tutorials are quite long, and I find them more difficult to read without it.

davepagurek:
[...]

PR #278, opened 2024-04-25
Default Astro syntax highlighting for non-embed Code Snippets

stalgiag:
This PR would add syntax highlighting back to the code snippets in tutorials. [Based on this conversation.](https://github.com/processing/p5.js-website/issues/221#issuecomment-2076069004)

Current:
Screenshot 2024-04-25 at 2 46 33 PM

After this PR:
Screenshot 2024-04-25 at 2 44 11 PM

The Figma mockups match the current but this sacrifices readability on longer snippets. The default Astro syntax highlighting could be replaced with syntax highlighting more similar to the `CodeEmbed` in the future but it does not make sense to attempt this before launch.

I'll leave it to the PF team to decide whether to merge.

### DIFF
--- a/src/components/CodeBlockWrapper/index.astro
+++ b/src/components/CodeBlockWrapper/index.astro
@@ -3,8 +3,7 @@ import CodeContainerWithCopy from "../CodeContainer/CodeContainerWithCopy.astro"
 ---
 
 <div>
-  <CodeContainerWithCopy>
-    <pre
-      class="[&_code]:p-0 [&_code]:bg-transparent [&_span]:!text-sidebar-type-color"><slot /></pre>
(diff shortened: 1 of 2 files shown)


davepagurek:
I'm in favour of this style for now, and potentially post-launch we can look into other highlighting themes (apparently [these are the built in options](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-themes)) that look closer to the original designs while still having highlighting. @Qianqianye what do you think?

Qianqianye:
Thanks @stalgiag. I agree with @davepagurek that we can go with the proposed style, and we can explore other highlighting themes after launch. I will merge this PR for now.

PR #286, opened 2024-04-26
Add paragraph on libraries page with submission link

davepagurek:
@Qianqianye was asking me about adding a bit of text to the libraries page explaining how to submit your own library.

I realized this would be a bit harder than just updating the subtitle string in the UI file because this ideally would be able to contain links. We were also thinking about adding an education resources page in the future, and this would likely run into similar issues, where we want translatable rich text to go on a single page.

I thought maybe the way to go would be to add a `pages` collection, whose content would get translated like e.g. tutorials, but that aren't directly attached to a section of the site like tutorials are. So for the libraries page, I'm manually pulling out the page that I want out of the `pages` collection via its slug. In the future, for a tutorials page, we can do something similar.

How do you all feel about this approach to adding rich text content to one-off pages?

Here's what the added paragraph looks like, mostly borrowed from the text on the current site: 
image

### DIFF
--- a/src/components/CodeBlockWrapper/index.astro
+++ b/src/components/CodeBlockWrapper/index.astro
@@ -3,8 +3,7 @@ import CodeContainerWithCopy from "../CodeContainer/CodeContainerWithCopy.astro"
 ---
 
 <div>
-  <CodeContainerWithCopy>
-    <pre
-      class="[&_code]:p-0 [&_code]:bg-transparent [&_span]:!text-sidebar-type-color"><slot /></pre>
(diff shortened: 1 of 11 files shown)


Qianqianye:
Thanks @davepagurek. Another approach could be add the paragraph about how to submit library in between the line break and first chapter title (eg. Drawing).

davepagurek:
Here's what it looks like now:
image

Every section has a divider between them though, so it does look a little bit like it's part of the Drawing section this way.

Qianqianye:
Thanks @davepagurek!

PR #288, opened 2024-04-27
Add education resources page

davepagurek:
@Qianqianye here's some scaffolding for an education resources page. For now I've based the layout on the tutorials page, let me know if something else makes sense! Especially if you want to just a more normal page layout, feel free to add more regular markdown in the first `<div>` and remove everything outside of it.

Some other notes:
- Currently this is based off of the changes in https://github.com/processing/p5.js-website/pull/286 for one-off pages, so that's the base for this PR until that's merged
- This is currently linked to from nowhere. Where do we want to link to it? Probably from community?
- This is also currently implemented as a single page rather than as a collection of multiple entries, because I'm not sure just yet what all we'll want to do with it in terms of ordering, layout, etc. @Qianqianye let me know if it would make more sense to get the content from a bunch of files similar to how libraries work, rather than this single-source version.

image


Qianqianye:
Thanks @davepagurek. I initially was thinking to link it in the bottom of Tutorial page. The Community page could also work. Which one do you think makes more sense?

davepagurek:
Update: added a link from the bottom of the tutorials page

image

Qianqianye:
Thanks @davepagurek!

PR #307, opened 2024-04-28
Add contents for the education resource page

tuantinghuang:
Added some resources from the doc, and added missing thumbnails to the folder too. Will try to continue adding the rest!


tuantinghuang:
I added all the contents in the google document to the page. After I ran a test build, I noticed a couple of issues:

- When they have multiple material links, the formatting would be a little off. I'm not sure how should I write in the markdown way that can give each item proper spacing or new line. (attaching screenshots of the issue)
Screenshot 2024-04-28 at 20 40 45
Screenshot 2024-04-28 at 20 41 13
<br><br>

- Some thumbnails look funny or pixelated after being cropped in the fix ratio (attaching two examples)

Screenshot 2024-04-28 at 20 45 37
Screenshot 2024-04-28 at 20 45 32

Qianqianye:
Thanks @tuantinghuang. I'm gonna merge this for now
Tagging @davepagurek to take a look at the multi material link issue.

davepagurek:
I think in Markdown you need a blank line between each item to get it to go onto its own line, or prefix each with `- ` to make a bullet list, but also we could turn it into a horizontal list of buttons if we think that will look better?

PR #313, opened 2024-04-29
Change styles for education resources materials links

davepagurek:
image

This makes the ones with multiple links look a tad better.


Qianqianye:
cc @tuantinghuang

tuantinghuang:
this looks great!

PR #336, opened 2024-05-06
More site details and fixes

davepagurek:
- In the reference page, subcategory titles are hidden when they are the same as the category title right above them
- Switches to using a light code theme:
  
  image

- Adds the ability for some components to expand to be more than the full width. Currently, the designs only account for a small fixed width, but some content includes multi column content (design and content were being worked on in parallel, so this is a case of not having the foresight for 100% of all the design requirements upfront.) An example of this is on the Intro to Shaders tutorial, which currently looks like this, making the code difficult to read:

  image

  This lets components optionally expand to use more than the original width, with the limitation that each *column* should have the same max wrap width for text to keep things readable:

  image

  Single-column content continues at the original width:

  image

  This also applies to editable code examples in tutorials, making them go side-by-side when there is room and when the sketch width is small (<= 200px wide) to match the reference page:

  image

- Changes the color of the p5 logo on hover (when normally black, magenta on hover; when normally magenta, black on hover)
- Adds UI translations for the titles in callouts ("Try This!", "Note", etc)
- Fixes some content bugs in tutorials (a google user content link by accident, a missing description)
- Removes the Related References header at the bottom of tutorials when the list is empty
- Fixes up some code styles in dark theme.

  Previously:

  image

  Now:

  image

  - It inverts the code blocks so that the now light theme blocks aren't super bright
  - It fixes the text color in inline code to make sure it has enough contrast
- Fixes a bug where tutorial titles were getting added to the `html` tag as classes. This was causing tutorials with "sketch" in the title (e.g. "optimizing your WebGL sketch") to render with the styles for a sketch page rather than a tutorial page.

@Qianqianye the remaining issues marked as p5.js internal are:
- "When the secondary side navigation panel vertically overflows, the title line should not stay fixed, it should scroll with the anchor link items. The only thing that should stay fixed is the dropdown arrow." I haven't tackled this one just yet since it's a little difficult making just one item sticky like that and it seems like a more minor detail. Will revisit at some point, probably.
- "Donate page’s iframe should have all rounded corners, not just the top left." I couldn't test this yet because I was working on an airplane and couldn't load that iframe 😅
- Any of the ones related to sketch previews being too small. Currently blocked on OpenProcessing for those.


limzykenneth:
@davepagurek Can you have a look at the merge conflict. Once that's resolved you can merge this. Thanks.

PR #345, opened 2024-05-08
Improving build performance

### LINKED ISSUE #340: Build optimization
Currently the build is taking a very long time with around 25-30+ minutes to complete. I'm looking into the causes and potential solutions. I have so far identified two large areas that are slow:
1. Reference pages generation
2. Asset optimization

For reference pages generation, I can track it all the way down to [this pair of loops](https://github.com/processing/p5.js-website/blob/main/src/pages/_utils.ts#L52-L62) which for each execution need to loop through up to (1023 * 1023) entries for each page (including translated reference pages). This on my M1 macbook pro measures to about 250ms per page which adds up to a lot.

The thing I noticed for this function is that it seems to be fully functional with no side effects, ie. the same arguments should give the same output. This means that it can be memoize. Indeed by memoizing the function with `lodash/memoize`, individual page's generation time goes down to ~20ms. To that end I would propose memoizing all of the fully functional functions which should greatly speed up build overall. If there's no obvious drawback on this I can start working on this.

For asset optimization I'm still investigating.

limzykenneth:
Fixes #340

Three of the most time consuming functions are now memoized using `lodash/memoize` which significantly pared down the build time for reference and sketches. For dev server the behavior does not seem to be affected as there seem to be a server restart on content changes, ie. updating content (including translation) will trigger a content refresh with updated content.

In addition, the image service used by Astro on build is switched to using `passthroughImageService` as it is doing duplicate task of `astro-compress`. Passing through speed things up slightly from a clean build and just let `astro-compress` do its thing.


limzykenneth:
@outofambit If you'd like to have a look at this to make sure things are not broken.

PR #347, opened 2024-05-08
keep mobile nav sticky on scroll

outofambit:


davepagurek:
Just a heads up, I think the `top: 80px` made the nav slide down farther than usual on desktop:

top-nav

outofambit:
@davepagurek thanks for catching that! fixed in 88c778b48e86585970b1efc1cfaf4573e93eb187

PR #353, opened 2024-05-15
Don't show search bar when clicking on an anchor link

davepagurek:
I wanted to take a crack at this to see if I could get it working to make the mobile experience a bit better (where just hovering at the top of the page isn't feasible.) My approach was to record when an anchor link is clicked, and detect when auto scrolling stops as a result of that, and prevent opening the settings bar in that in between phase. I tried to make it robust to any dynamically added anchors with a mutation observer, not sure if that's overkill or not.

anchor-scroll


davepagurek:
Thanks for testing it out!

PR #361, opened 2024-05-22
Add banner

davepagurek:
(Ignore the astro dev toolbar in screenshots)

It sticks to the bottom of the page, has a message, and can be closed:

banner

It takes on the accent color of the page:

image

When the user dismisses the banner, it stores the `title` we gave that message in the browser's local storage. That way, we can only show the banner again once we have a new message with a new `title`.

Also, when testing this out, I noticed some issues with scroll bars appearing on the homepage due to the header and hero image not taking into account scrollbar width. I've corrected those here too.


davepagurek:
oh that's a good callout, thanks! I also have a content/pages directory that is being used for similar things because I didn't know about the text-detail directory at the time, would that avoid the issue of getting a page made for it? if so maybe I should keep it around and use it just for these non routed bits?

outofambit:
@davepagurek maybe? it looks like Education resources would work great as a page in text-detail, I'm not sure where the libraries file in there is used. Actually, looking more at the metadata you have for the banner, I would suggest that it should be its own collection. Nothing else in the text-detail collection or in content/pages collection is going to have the same metadata as a banner and i think it would keep the schemas a lot clearer to keep them separate.

davepagurek:
I updated `banner` to be its own collection now!

outofambit (commented):
Looks great, just one more thing! New collections should be also imported and added to the config in `src/content/config.ts` (see https://docs.astro.build/en/guides/content-collections/#defining-multiple-collections for an example of this pattern). It does seem like its still working without that though which I don't understand.

davepagurek:
Oops good catch, thanks!

outofambit (approved):

Qianqianye:
We will merge this after we hear confirmation from the designers. @stefanietam @tyler-yin

tyler-yin:
Looks good to us, thanks!

Issue #359, opened 2024-05-21
Potential 1.9.4 release this week


Qianqianye:
We updated more  reference last couple of weeks. We'd like to do a patch `1.9.4` release on this Tuesday or Wednesday, and we want to double check with you first to see if it's okay. Thanks!  @outofambit @stalgiag

CC @limzykenneth @davepagurek

outofambit:
@Qianqianye yep, sounds good! The relevant steps to update the website after a p5 release are in https://github.com/processing/p5.js-website/blob/main/docs/scripts.md. I will add a few more details to that page shortly if someone else would like to give it a try this time?

limzykenneth:
I had a look at the possible steps for release here and there are two things that need to be hooked up, both relating to version numbers.

We want version numbers on website for anywhere that references it and relevant CDN script tags to be updated to the latest when there is a release. The first place is in the "Download" page where I see that the number is currently hard coded in. The second in for the embeds which is used by CDN links is defined as a global value which is also hardcoded and it is in TS so would be trickier to script an automated number update, as oppose to having a JSON file that can be easily parsed and updated.

The YUIDoc JSON file should have a version number in there already but it is a bit more about simple build still having the version number without having to fetch the library website for it (or anything online really).

PR #364, opened 2024-05-22
add script to pull in latest p5 version

outofambit:
and use across website pages

re: https://github.com/processing/p5.js-website/issues/359#issuecomment-2125677066

### DIFF
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -18,7 +18,11 @@ To update the content on the p5.js website following a new release of p5.js, you
 
 ## Contributors List Build Script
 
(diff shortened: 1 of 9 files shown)


limzykenneth (approved):

Qianqianye:
While I'm checking the reference page, I noticed that some of the reference are not showing the updates from the latest [1.9.4 release](https://github.com/processing/p5.js/releases/tag/v1.9.4). For example, [draw()](https://p5js.org/reference/#/p5/draw) in the old p5.js website is updated [last month](https://github.com/processing/p5.js/blob/v1.9.4/src/core/main.js#L176), but [draw()](https://beta.p5js.org/reference/p5/draw/) in the new website is updated [three months ago](https://github.com/processing/p5.js-website/blob/main/src/content/reference/en/p5/draw.mdx?plain=1). 

Can we double check if the new reference from the 1.9.4 release are properly built in the new website? @outofambit @davepagurek @limzykenneth @ccanash

limzykenneth:
We need to manually run the script to update the reference before committing and pushing it. I'll make that part of the release action after we have launched fully.

Issue #367, opened 2024-05-23
Reference page spacing not correct


Qianqianye:
The spacing below the heaers are not correct in a few sections on reference page. See some screenshots below.

CC @outofambit @davepagurek @limzykenneth @ccanash 

<img width="1237" alt="image" src="https://github.com/processing/p5.js-website/assets/18587130/6237c004-95f6-461c-8643-7c1019421724">

<img width="1228" alt="image" src="https://github.com/processing/p5.js-website/assets/18587130/62e46600-efc8-4cde-8405-d61f7aa8ab48">

davepagurek:
I bet these are the elements that don't have a subcategory because the subcategory would have had a duplicate name as the category. I can take a look tonight!

PR #370, opened 2024-05-23
Fix gap under reference section headings

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/367

This adds back the margin under the category headings if there will be no subcategory heading right underneath.

image


Qianqianye:
Look great, thanks @davepagurek!

Issue #371, opened 2024-05-23
Alt text not showing on all event pages


Qianqianye:
Image Alt text not showing on all event pages, except first image. See a screenshot below:

CC @outofambit @davepagurek @limzykenneth @ccanash

<img width="788" alt="image" src="https://github.com/processing/p5.js-website/assets/18587130/ad493313-bd71-42d5-a2a3-a54e6f4f5ec6">

PR #372, opened 2024-05-23
Display images with alts on event pages

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/371

This passes in the same image component used for tutorials for the images on event pages, which renders the alt text as a viewable element.

image


Qianqianye:
Thanks @davepagurek!

Issue #376, opened 2024-05-23
Alt text not showing on homepage hero images


Qianqianye:
We have included image description and alt text for all the images. Alt text is not showing on homepage hero images when 'View Alt Text' is checked. See screenshot below:

CC @outofambit @davepagurek @limzykenneth @ccanash

<img width="1052" alt="image" src="https://github.com/processing/p5.js-website/assets/18587130/38076cd0-16a5-4811-a7f6-0555881a25c6">

PR #377, opened 2024-05-24
Show alt text on homepage images

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/376

I was initially going to use the Image component in the components folder, but it seemed to not like the string `src`s we were passing into the `` tag, and also I needed to add some classes to the container itself to make it show/hide everything correctly, so in the end I just mirrored some of the same structure in the HomePage component.

image


Qianqianye:
Thanks @davepagurek!

PR #378, opened 2024-05-24
Round both top corners of the donate box

davepagurek:
We can't round the bottom two corners because we can't style the inner content of the `<iframe>` and also the height changes on each step of the donation flow, but since the container constraints the width enough to make the header always full-width of the container, moving the `border-radius` from the container to the iframe itself seems to get both top corners to be round.

image


Qianqianye:
Thanks @davepagurek!

PR #381, opened 2024-05-24
Attempt to use Image component instead of img on homepage

davepagurek:
Some homepage images don't seem to be optimized by Astro. I'm not sure yet exactly why this is, but maybe using an astro `Image` component instead of `` will help?


Qianqianye:
Thanks @davepagurek! While this resolves the issue on desktop, the image loading process is a bit slow on mobile.

Wondering if @outofambit or @limzykenneth have any thoughts on this.

PR #382, opened 2024-05-24
Restore example history

calebfoss:
Restore git history to examples, which was removed in [this commit](https://github.com/processing/p5.js-website/commit/37e3d669bfd71876b3c75341e11573d4b764c64c).


limzykenneth:
Actually looking into it now, I don't think the history is restored? At least it is not associated with the example file's commit history?

calebfoss:
@limzykenneth I am seeing that too. On my local clone of my forked repo, the history is restored, but it looks like that did not carry over once I pushed to Github.

To answer your first question, I merged in the example repo with allow-unrelated-histories, removed the files with missing history and committed that, moved the files with history to their corresponding location and committed that, and then removed some additional files from the example repo.

I wonder if that process needs to happen on the original repo, rather than a fork.

calebfoss:
Oddly, all the commits from that history are now part of the repo's history, and you can see that reflected in the contributors list. The blame for individual files, though, remains unchanged.

calebfoss:
Maybe using the [rebase & merge](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#rebasing-and-merging-your-commits) option would fix this?

calebfoss:
I can see the history on this repo if I look at the previous commit:  https://github.com/processing/p5.js-website/tree/d6b28924185dac5e25c39cd0fc2fc4a1f6392883/src/content/examples/en
So it looks like the merge commit is what reverted it back. 

Pardon all the individual messages! I keep noticing more details.

davepagurek:
We can give rebase and merge a shot, do we have to revert this PR first and make a new one? or alternatively we could just take your branch, apply it to main, and force push over it -- not good practice I guess but we won't be making a habit of it.

calebfoss:
Honestly this is deeper into git than I'm familiar with, so my confidence here is low: I think that, yes, it makes sense to revert the PR merge first before reopening to rebase & merge. If that doesn't work, your second suggestion makes sense, but I figure it makes sense, as you referenced, to ideally avoid force rewriting history on the main branch. 

Thank y'all for helping with these complications!

davepagurek:
I merged the revert PR, do we now need to do something to the commits in order to re-apply them? since the revert just added a revert commit, all the old ones are still in the history.

calebfoss:
Thanks! Unfortunately, I'm not finding a way to reopen the PR, since now Github sees the main branch as being up to date with mine. So I think this may require a forced change to the main branch. One route would be to rebase the main branch and replace the commit that added the example files without history with one that adds them with history. I just tried working on that on my fork, but I realized that I don't think you could merge a PR with rebased history?

PR #387, opened 2024-05-29
Created SPANISH setup.mdx

lauradowell:
Added a Spanish translated version of the setup() function. I also made sure that "Structure" module andf submodule is referred to as "Structure" In other Spanish translations on the website


limzykenneth:
@lauradowell This seems to be a duplicate of #388?

Issue #390, opened 2024-05-29
The banner won't appear on a new tab after you close it once


Qianqianye:
The banner won't appear on a new tab after you close it once. Seems like an issue with sessionStorage and localStorage.

cc @davepagurek @limzykenneth

limzykenneth:
Specifically we can use sessionStorage instead of localStorage to store the flag on whether to hide the banner or not so that it will still show up on new browser sessions.

davepagurek:
Ah ok, I wasn't sure how permanently we wanted it hidden, but yeah this will be a quick change!

PR #392, opened 2024-05-30
Hide banner for session only instead of permanently

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/390


Qianqianye:
Thanks @davepagurek!

Issue #405, opened 2024-06-11
Code in reference not appearing correctly


davepagurek:
In the reference, it looks like we have some code blocks just placed within `<code>`, e.g. this bit in the `orbitControl` reference:

```
 * The fourth parameter, `options`, is also optional. It’s an object that
 * changes the behavior of orbiting. For example, calling
 * `orbitControl(1, 1, 1, options)` keeps the default sensitivity values while
 * changing the behaviors set with `options`. The object can have the
 * following properties:
 *
 * <code>
 * let options = {
 *   // Setting this to false makes mobile interactions smoother by
 *   // preventing accidental interactions with the page while orbiting.
 *   // By default, it's true.
 *   disableTouchActions: true,
 *
 *   // Setting this to true makes the camera always rotate in the
 *   // direction the mouse/touch is moving.
 *   // By default, it's false.
 *   freeRotation: false
 * };
 *
 * orbitControl(1, 1, 1, options);
 * </code>
```

This gets incorrectly rendered as *inline* code, with some `<p>`s in there too:

![image](https://github.com/processing/p5.js-website/assets/5315059/67ff920e-857f-4668-a9b2-a14cf1bcc8f6)

davepagurek:
hmm this is actually broken in the live p5.js site too:
![image](https://github.com/processing/p5.js-website/assets/5315059/5a58b7b5-2bd4-4270-9909-7aa95e16647d)

davepagurek:
Oh this is fixed in https://github.com/processing/p5.js/pull/7089. I guess we aren't going to immediately fix this in a new release, so maybe for now I'll manually update the references in this repo

Qianqianye:
Adding @nickmcintyre to this discussion.

PR #407, opened 2024-06-12
Update some code blocks

davepagurek:
Fixes https://github.com/processing/p5.js-website/issues/405.

**What I've done so far:**
- [x] Manually cherry-pick https://github.com/processing/p5.js/pull/7089 and https://github.com/processing/p5.js/pull/7092 onto the `v1.9.4` tag before running `npm run build:reference`
- [x] Fix rendering of code blocks


davepagurek:
@Qianqianye there are some technical challenges to getting reference items to display as highlighted code blocks since, rather than using the astro mdx format everything else uses, it stores raw html in a string in its source files. For now I've opted for a CSS-only solution that at least renders them as blocks.

I think this is probably good enough for now, let me know what you think though!

## Before:
image


## After:
image

PR #428, opened 2024-07-01
Update h4 styles

davepagurek:
<table>
<tr>
<th>Before</th><th>After</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

Other spots that used h4 have been updated to look the same as before (and it seems to be just these two cases according to this search https://github.com/search?q=repo%3Aprocessing%2Fp5.js-website%20h4&type=code):
- The param name in the list of parameters in a reference page:

  image
- The author on education resources entries:
  image


Qianqianye:
Thanks @davepagurek!

PR #438, opened 2024-07-05
Add directory for all libraries

davepagurek:
This separates the libraries page into two pages: `/libraries`, which shows full-sized previews and up to 4 entries per category, and `/libraries/directory`, which has a list view with everything.

The idea is that the main page for libraries gives a more randomized view where we don't feature any individual contributor too many times, while still providing a view where we can show everything and encourage people to contribute as much as they want.

The main libraries page gets a Featured / Everything toggle at the top, and a View All button at the bottom of each section.

image

Clicking Everything or a View All link goes to the directories page, which has a list layout. The View All button for a section links directly to the start of the list for that category.

image

image


Qianqianye:
Thanks @davepagurek. I'm wondering if the website designers @stefanietam @tyler-yin have any insights on the new design.

Issue #433, opened 2024-07-03
Examples have play/stop buttons cut off


davepagurek:
<img width="1149" alt="image" src="https://github.com/processing/p5.js-website/assets/5315059/d23b6f0e-bc06-493f-87b3-e91af2b4a0f8">

I think my update to size community sketches to fit the window may have broken this?

Qianqianye:
Thanks @davepagurek. Tagging @outofambit @stalgiag as they might be able to offer insights.

wagedu:
Not sure if it's related but some of the Reference examples are auto-run while others need the Play button to be pressed.
ie:
https://p5js.org/reference/p5/arc/
3 first examples auto-run, next 7 don't

https://p5js.org/reference/p5/ellipse/
4 first examples auto-run, the 5th don't

davepagurek:
Separate issue I think, I'll make a new issue for this

PR #440, opened 2024-07-05
Fixes for examples layouts

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/433

So this ended up being a little bigger than I thought it would. So the original problem was that:
- Most of the time, examples are 710px wide (I noticed a few 720px ones too though)
- The play/stop buttons are 40px wide + 20px padding, so the total example width is 770px
- The example page layout was only 750px wide

The 750 number confused me a bit because a lot of the other layouts are 767px wide. Even if we make the examples page match the other 767px layouts, though, it would still clip the play/stop buttons because it needs 770px to display everything fully.

So the solution I came to for now is:
- Use 770px everywhere
- Add some code to place the play/stop controls below the sketch when there isn't 60px of room before the breakpoint width

Some alternate solutions could be to adapt the examples to use a narrower width (I opted for the one changing fewer files for now), or to change no layouts and just have the controls go below the sketch for 710px-wide examples.

...and then finally I also noticed that the example I had originally been looking at, https://p5js.org/examples/imported-media-video-canvas/, wasn't actually showing any video due to it using the URL scheme `assets/` rather than `/assets/`. So I've additionally corrected the cases that don't use the preceding slash.


Qianqianye:
Thanks @davepagurek. Could you add a screenshot of the new example page layout you propose? I'm wondering if the website designers @stefanietam @tyler-yin have any insights on the new design. cc @limzykenneth 

For reference, below is the designed example page
image

davepagurek:
Here's a gif tabbing back and forth between this branch and the live site (the grey canvas is live, the one with the actual video shown is this branch.)

The layout overall is the same, the breakpoint is just a tad wider:

examples-before-after

It's even subtler on the sketches page, a 3px difference (ignore the header appearing/disappearing, that seems to be because of the domain hosting the site + the OP API):

sketches-before-after

Qianqianye:
Thanks @davepagurek! Let's merge this for now after the conflicts are resolved!

PR #470, opened 2024-07-26
Fixes issue #469: storeItem value changed from any type to serializable

SembeiNorimaki:
storeItem value changed from any type to serializable.


limzykenneth:
As @Qianqianye for the changes to the reference to persist and not be overwritten, it needs to be updated in the inline documentation here: https://github.com/processing/p5.js/blob/main/src/data/local_storage.js#L25. I think the language can be changed but "serializable" for me is just a bit too jargon for newer users of p5.js. Being a bit more explicit around accepted types may be better.

In any case I will close this for now. Please open a relevant issue for PR to discuss this on the library repo where we can accept the changes. Thanks @SembeiNorimaki.

Issue #458, opened 2024-07-13
Expected search results missing under Reference section


meezwhite:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

None

### Actual Behavior

I was searching for "color" or "color()" using the website search.

Under the "Reference" section, the first suggestion is "color". However, this links to [p5.Framebuffer/color](https://p5js.org/reference/p5.Framebuffer/color), but there is no result linking to the [color() reference page](https://p5js.org/reference/p5/color/).

### Expected Behavior

I was expecting "color" to link to the [color() reference page](https://p5js.org/reference/p5/color/).

### Steps to reproduce

1. Search for "color" or "color()" using the website search, or follow this link: https://p5js.org/search/?term=color
2. Click on "color" under "Reference" section.

### Would you like to work on the issue?

Unsure

meezwhite:
This issue also affects other search queries e.g.:
* `get` or `get()` shows result for [p5.TypedDict/get](https://p5js.org/reference/p5.TypedDict/get), however no result for [pixels `get()` function](https://p5js.org/reference/p5/get)
* `set` or `set()` shows result for [p5.Vector/set/](https://p5js.org/reference/p5.Vector/set), however no results for [pixels `get()` function](https://p5js.org/reference/p5/set) or [p5.TypedDict/set function](https://p5js.org/reference/p5.TypedDict/set)

Qianqianye:
Thanks @meezwhite! 
I'm tagging the web developer consultants @outofambit @stalgiag to take a look at this. The global search is a bit buggy when it comes to entries that have same names (but in different places), like color, get, set, mousePressed, as mentioned in earlier part of this issue. Could you share some insights on this? Thank you!

cc @limzykenneth

stalgiag:
Ah that makes sense! It looks like the script for parsing the search index uses [the title from the content](https://github.com/processing/p5.js-website/blob/main/src/scripts/builders/search.ts#L238) when building the reference search index. The title in the reference content items does not include the class name as the standalone method name is used on the Reference index page and that item's status as a member method is derived from the content file location rather than the title. As a result, the search index will overwrite any previous references with that same title in [this line](https://github.com/processing/p5.js-website/blob/main/src/scripts/builders/search.ts#L269). The to-do would be to include the class name in the title for the reference items. There are several approaches to getting the class name from the file item in that script.

[...]

PR #472, opened 2024-07-28
Fix reference search index

stalgiag:
addresses #458 

This PR fixes an issue where methods with the same name were overwriting each other in the search index due to ignoring class names. This PR does the following:

- Adds method class names into relevant reference search indices
- Adds parentheses to methods in the search index for a more consistent appearance across the site
- Adds aliases to reference items that have a class or parentheses so they still rank high in searches without those elements included
- Adds a set of unit tests that ensures the above

~~Will request review after CI build finishes~~


stalgiag:
Also, I was hoping to add @meezwhite as a reviewer but the repository settings don't allow review requests from people without a maintainer role.

I requested a review from three maintainers just as a way of flagging. One reviewer is likely enough. Thanks!

davepagurek (approved):
This looks good to me, thanks @stalgiag!

Qianqianye:
Thanks @stalgiag!

meezwhite:
Looks great! Thanks @stalgiag!

PR #481, opened 2024-08-08
Fixed inline code rendering issue in `curveVertex`

DenisovichDev:
Hello, this is my first PR here!

I noticed some inline code issues that might have popped up because of some porting issues from the old website. This is what it looked like:
image

Here's what it looks like now:
image

These were some special cases in the reference where it wasn't a code snippet but a multiline inline codes.

### DIFF
--- a/src/content/reference/en/p5/curveVertex.mdx
+++ b/src/content/reference/en/p5/curveVertex.mdx
@@ -31,17 +31,28 @@ description: >
 
   <a href="/reference/p5/endShape">endShape()</a> in order to draw a curve:</p>
 
-  <p>
-  <code>beginShape();</code><br>
-  <code>// Add the first control point.</code><br>
-  <code>curveVertex(84, 91);</code><br>
-  <code>// Add the anchor points to draw between.</code><br>
-  <code>curveVertex(68, 19);</code><br>
-  <code>curveVertex(21, 17);</code><br>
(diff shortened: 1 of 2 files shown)


davepagurek:
Thanks @DenisovichDev! The reference gets updated every release from the contents in the p5.js repo, so I think we'll need to update the source there to not get turned back into multiple lines. I fixed a similar issue in the p5.js repo here https://github.com/processing/p5.js/pull/7089/files, does that method also apply to these functions?

Within this repo, I think rather than having inline code blocks separated by newlines, we might want to just group everything into one code block. That ends up looking like this: https://github.com/processing/p5.js-website/pull/407/files#diff-a9f22d63312d5dcf8286875d13c5ca83ee171d6bb92a2a21b8c50b7090771010R55

DenisovichDev:
Thanks for the reply! I suppose that would fix it. So, shall I make a pull request changing the docstring in the p5.js repo?

And for this repo, I'll make the changes that you mentioned then, by grouping it with the \<pre> tags

davepagurek:
Sounds good, thanks so much!

DenisovichDev:
While editing the reference from p5.js repo I noticed a similar issue in `normal()` as well, so I have made the changes and added it here

davepagurek (approved):
Thanks for all the fixes @DenisovichDev!

PR #486, opened 2024-08-14
Correct some Korean translations

cog25:
I fixed some korean translation.

Some titles are wrong. They might be name of function and so shouldn't be translated. 
Also, `itemtype` in frontmatter shouldn't be translated.


Qianqianye:
Thanks @cog25! I'm tagging the Korean translation stewards @yinhwa @designerSejinOH to take a look at this.

yinhwa (approved):
Hi @cog25 thank you very much for finding & fixing this! @Qianqianye @designerSejinOH the commit looks good to be merged :)

PR #488, opened 2024-08-15
Apply /reference path replacements to search index build scripts.

### LINKED ISSUE #485: 404 error when looking up references for WEBGL, WEBGL2

Other (specify if possible)

Mac OS

### Web browser and version

Safari Version 17.6 (19618.3.11.11.5)

### Actual Behavior

when you use the search function to look up references or go through the reference list for either of the two, and on links to examples, you reach a 404 page.

https://github.com/user-attachments/assets/8eb3fed4-eaba-4bae-82a5-10fab8aeb93b

### Expected Behavior

well, the reference page

### Steps to reproduce

_No response_

no

mayaarguelles:
This resolves #485, which was caused by the paths of items in subfolders of the reference directory being modified in the Astro router but not in the search index build scripts generating the relative URLs to each search result. To address this, I've moved the string replacement previously happening in [normalizeReferenceRoute](https://github.com/processing/p5.js-website/blob/15ff0b20d7257c95fc90aa6e274e215f72ae17b1/src/pages/_utils.ts#L154) to a separate function which I then used in the search index build scripts to apply the same transformation. 

One small caveat — I had trouble importing any function from the [pages/_utils.ts](https://github.com/processing/p5.js-website/blob/main/src/pages/_utils.ts) file in the search build script due to the astro:content dependencies included in that file, so I moved this function to the [pages/_utils-node.ts](https://github.com/processing/p5.js-website/blob/main/src/pages/_utils-node.ts) file, which is probably not a huge deal but maybe worth noting as it doesn't technically rely on any runtime node dependencies. With the function in the [pages/_utils.ts](https://github.com/processing/p5.js-website/blob/main/src/pages/_utils.ts) file the build would work perfectly through the total `npm run build`/`astro build`, but fail on the independent `npm run build:search` with the message:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data are supported by the default ESM loader. Received protocol 'astro:'
```

Moving the function to the other utils file solved this, and this standardization solves the issue with the links to any items from the constants or types section 404ing, but maybe I'm missing an easier way that would let that function live in the same file it's getting used for the router and still be imported to the search builder.


davepagurek (approved):
Thanks!

PR #499, opened 2024-08-30
Fix double-rendering of libraries page description

davepagurek:
This is currently branching off of #500 since that fixes the tests on main, that one should be merged before this.

Previously this was done manually, but now this happens automatically when there is a matching page in `src/content/pages`.

image


Qianqianye (approved):

Qianqianye:
Thanks @davepagurek!

PR #506, opened 2024-09-04
Added top margin for h2 tag

shourysingh07:
**Added a top margin to the h2 tag in global.scss, which also fixed the other page's headings that still had issues.**

### **Before**

before

### **After**

after


Qianqianye:
Thanks @shourysingh07!

PR #521, opened 2024-09-07
Fix tables and lists in reference items

davepagurek:
I was looking into making a custom build of the site to show shader hooks docs on for people testing out the feature before it's merged and deployed.

I noticed a few little styling issues in reference pages:
- Lists don't have numbers or bullets
- Table cells aren't aligned to the top
image

This addresses those issues:
image


Qianqianye:
Thanks @davepagurek!

Issue #522, opened 2024-09-07
Problem with html symbols at https://p5js.org/reference/#Foundation


goldwasser:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Mac OS

### Web browser and version

128.0.6613.120 (Official Build) (x86_64) 

### Actual Behavior

For both the > and >= summary description as visible on https://p5js.org/reference/#Foundation errant symbols are being shown as seen in image below. 
<img width="303" alt="Screenshot 2024-09-07 at 9 51 55 AM" src="https://github.com/user-attachments/assets/3a8a4a4b-d930-4da5-851e-4297d7e496a1">

### Expected Behavior

If mirroring approach that seems to happen for < and <= the symbols can just be omitted so that summary reads "The greater than or equal to operator evaluates to true if ..."

### Would you like to work on the issue?

no

meganmckissack:
I'd like to work on this, I found a few other related bugs in the same foundation section of the reference, all related to the page and url name changes from `<` type symbols characters like "lt".

davepagurek:
I left some discussion on a related issue here https://github.com/processing/p5.js-website/issues/447#issuecomment-2239528853 about how to diagnose the source of the issues, so that you can target a fix in the right spot. Maybe take a look and see where the same reasoning leads you for this one.

davepagurek:
Update on this: turns out those operator references don't even exist any more in the p5.js repo! I thought I was going crazy haha, but I think the problem is that when we build the reference, it just creates/modifies new files, but never deletes old ones if items are removed from the reference.

Issue #447, opened 2024-07-07
Some symbols in code blocks in the reference (the description, not the examples) are msitakenly HTML-escaped


davepagurek:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Mac OS

### Web browser and version

Firefox

### Actual Behavior

On https://p5js.org/reference/p5/mousePressed/, it contains this code snippet:

```
if (mouseX &lt; 50) {
  // Code to run if the mouse is on the left.
}
```

### Expected Behavior

```
if (mouseX < 50) {
  // Code to run if the mouse is on the left.
}
```

### Steps to reproduce

Go to https://p5js.org/reference/p5/mousePressed/

### Would you like to work on the issue?

Need to discuss the fix first

Abhinavcode13:
```
if (mouseX &lt; 50) {
  // Code to run if the mouse is on the left.
}
```

don't this work fine ?

davepagurek:
When users write code in the p5 web editor, they should be writing `<`, not `&lt;`, so the `&lt;` should not end up visible to users in the docs. Behind the scenes for this website it may still need to be represented like that, but we should make sure it gets rendered as the HTML character it represents instead.

Abhinavcode13:
so the most feasible fix should be to replace the `&lt` with `<` ?

davepagurek:
It starts as `<` in the source of the documentation here:

https://github.com/processing/p5.js/blob/db01247bd224b71fbf253a546f6a35e65b988267/src/events/mouse.js#L1101-L1104

It's also still a `<` after it gets imported into the website repo, so the import script isn't the problem:
https://github.com/processing/p5.js-website/blob/798342779820ea53d671cf85bf6b5731fb6f3c23/src/content/reference/en/p5/mousePressed.mdx?plain=1#L28-L30

It looks like we're intentionally escaping the contents of `<code>` blocks here: https://github.com/processing/p5.js-website/blob/798342779820ea53d671cf85bf6b5731fb6f3c23/src/pages/_utils.ts#L230-L244

Looking at the git history, this was to address https://github.com/processing/p5.js-website/issues/95, which is for *inline* code. I think what we want is to have this apply just to *block* code. So that would mean escaping `<code>` tag contents, but *not* the contents of `<pre><code>` I think? And we would need to do that in the escape function I referenced above.

Qianqianye:
[...]

PR #525, opened 2024-09-08
Fix url paths and inline html character bugs on Foundation section of reference page content #522

meganmckissack:
Addresses #522

Some of the summary descriptions show double characters when using operators like > and >=.  Also, when clicking through to the full-page reference for these operators, the internal reference urls direct to 404 error pages since they are using the operator symbols rather than the new page url initials.

_Character doubling_ 
**Before**
Screenshot 2024-09-07 at 5 39 16 PM
Screenshot 2024-09-07 at 5 51 44 PM

**After**
Screenshot 2024-09-07 at 5 41 46 PM
Screenshot 2024-09-07 at 5 52 47 PM

_URL change_
**Before**
Screenshot 2024-09-07 at 5 55 25 PM

**After**
Screenshot 2024-09-07 at 5 56 30 PM


meganmckissack:
After opening this pull request, [this comment](https://github.com/processing/p5.js-website/issues/522#issuecomment-2336481355) on #522 mentions that the character doubling may be coming from a different source that I'll look into.

The failing tests are mention the same `{rewriteRelativeLink}` import referenced in issue #524.  I don't know if it is what's causing the failed tests or because I manually changed the url paths in the mdx files yet.

Qianqianye:
Hi @meganmckissack, thanks for opening this PR. The reference pages are built directly from the p5.js library [src](https://github.com/processing/p5.js/tree/main/src) folder, so we will need to edit the source code directly to fix the links.

I noticed that there might be some bugs rendering and showing the source code links (at the bottom of the reference pages) within the following pages:
- [>](https://p5js.org/reference/p5/gt/) (showing as '&gt'), 
- [>=](https://p5js.org/reference/p5/gte/)(showing as '&gt;='), 
- [<](https://p5js.org/reference/p5/lt/) (showing as '&lt'), 
- [<=](https://p5js.org/reference/p5/lte/) (showing as '&lt;='), 

@nickmcintyre @limzykenneth @davepagurek, do you have any thoughts on this? Thank you.

limzykenneth:
For rendering in the page I'm not sure if Astro or the current build automatically encodes the characters so we need to check there first. For URL, adding a step to encode the URL path may be enough to solve this without needing to change the reference in the main p5.js repo.

Qianqianye:
Thanks @limzykenneth.
This PR is related to issue #538 and  https://github.com/processing/p5.js-website/issues/447#issuecomment-2239528853, which are about rendering symbols and characters (for example "<>" ) on reference page. To keep the discussion organized and consolidated, let's continue the discussion within issue https://github.com/processing/p5.js-website/issues/447. Thank you!

Issue #539, opened 2024-09-12
Code blocks appear on single line


skkdevcraft:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Windows

### Web browser and version

Chrome 128.0.6613.121 (Official Build) (64-bit) (cohort: Control); Firefox 130.0

### Actual Behavior

Code blocks appear on single line.

Current behavior:
![image](https://github.com/user-attachments/assets/67b7564f-0e20-48e9-86c5-c12cd425eb3c)

### Expected Behavior

Expected behavior:
![image](https://github.com/user-attachments/assets/83cb4ddd-1a3d-45bd-92e6-3dfe79065695)

### Steps to reproduce

Navigate to: https://p5js.org/contribute/contributing_to_the_p5js_reference/

### Would you like to work on the issue?

yes

skkdevcraft:
Appears to be fixed by #536

PR #540, opened 2024-09-12
Allow markdown code blocks to show new-lines

skkdevcraft:
Fix for #539


skkdevcraft:
There is an alternative fix already applied.

PR #544, opened 2024-09-14
Handle @beta in reference items

davepagurek:
This adds support for a `@beta` tag in our jsdocs to label a feature as experimental.

There aren't any of these on main yet, but I'm considering using it for shader hooks in https://github.com/processing/p5.js/pull/7149. Here's how it would look for that:

On the reference listing, beta APIs get a flask icon:
image

On an individual item's reference, it gets a little explanation on the top:
image


Qianqianye:
Thanks @davepagurek. How do you feel about changing the highlighted background color to the reference page theme yellow color #deed33?

davepagurek:
Sounds good, updated! Here's what that looks like on an arbitrary reference item as an example:

image

Qianqianye:
Looks great, thanks @davepagurek!

Issue #473, opened 2024-07-29
Problems with antialias / Blurred text


Nemecsek:
### Most appropriate sections of the p5.js website?

Other (specify if possible)

### What is your operating system?

Linux

### Web browser and version

Chrome 124.0, Firefox 128.0, Brave 1.16.76

### Actual Behavior

Hi. 
I like the organization of the new P5js.org site, but I see it "blurred". Jumping to the old one the text is clean and perfectly readable. I tried with multiple browsers (Chrome, Firefox, Brave, Edge) on both Ubuntu 20 and Windows 11, and also on different monitors.

![blur](https://github.com/user-attachments/assets/fa4541a6-9f97-4f68-aa7a-ee7d228f1310)

There is something off, like the antialias not working properly: Characters shows spurious artifacts (shadows) and give a bad look to the page.

Is it due to a poor font choice or to my hardware/settings?

### Expected Behavior

Possibly change the font.

### Would you like to work on the issue?

I don't have the skills to work on this problem.

Nemecsek:
Ubuntu 20.04.6 LTS, Gnome 3.36.8, X11.
I also tried on Windows 11 and the effect is the same.

limzykenneth:
Ok, that is odd, I can vaguely see this on my machine as well but only when I view it on a 1440p monitor and not on my 4K monitor. @Nemecsek Can you perhaps also provide your screen resolution as well?

Nemecsek:
Thank you for the time you are taking to analyse this issue. 

I have a Fujitsu Celsius laptop with native resolution of 1920x1080 (16:9). 
I connected a couple of Benq monitors, both 1920x1080 too. I also tried on Windows on the same laptop model. 
In all my systems the issue is quite visible even if in screen snapshots it is not so obvious. 

I am not a font expert but I tried with other fonts to check what's going on, editing the html in the original page. 

My observations:
1) The original font (National Park) is 18.75px. The fractional size introduces artifacts that are not so visible when we use a "standard" 16px. Some characters, such as lowercase **e** and **a**, are in this font "closed" and this introduces unpleasant shadows after rendering due to antialias.

2) The font is so big, and I was zooming the page to 70% without thinking about it (sorry, I missed this fact as it is done not consciously). This action lowers text quality.

3) Some fonts are definitely better than National Park (all 16px to avoid the page zoom). I personally find Calibri and Garamond 16px a lot more readable. Alas Github blurs my screnshot:
   
[...]

PR #545, opened 2024-09-14
Fix text stroke color not the same as the fill color

### LINKED ISSUE #518: Jagged Hyperlink Text (Dark Mode)

Other (specify if possible)
Any page on the new p5.js website

Windows 11 Pro 10.0.22631, MacOS Ventura 13.3.1, Android 14 (Tested on multiple devices + browsers)

### Web browser and version

Chrome 128.0 (on Windows and Android), Firefox 130.0, Microsoft Edge 128.0, Chrome 123.0 (on MacOS)

### Actual Behavior

Heyo!

I was just trying out the new p5.js site in dark mode and noticed the typography (particularly the magenta text) was looking super jagged/aliased. on some screens, regardless of browser. Magenta hyperlink text has a thin outline on high PPI screens (Mobile and MacOS Retina) so the source of the aliasing is most likely that. The outline is the same color as the base text. Tested in light mode and the issue is there but not as noticeable. My guess is that this could also be linked to [this issue](https://github.com/processing/p5.js-website/issues/473) someone else brought up about blurred text.

Screenshot of jagged text on Chrome browser on Windows device with low PPI screen

_^ This is what the text looks like on Chrome on a Windows device_

Screenshot_2024-09-06_at_7 29 45_PM

_^ This is what the text looks like on Chrome on a MacOS device_

Screenshot_2024-09-06_at_7 29 45_PM

_^ This is what the text looks like on Chrome on a Android 14 phone_

### Expected Behavior

Not sure if this behavior is intended/expected as it does appear on other pages as well. It does make the typography look worse on low PPI screens and overall harder to read.

### Steps to reproduce

Navigate to any page with the magenta-colored hyperlink text. Effect worsens with lower PPI screens. 

Sure, I'd be happy to work on the issue! 😁 Just identify which solution (see below) would be desired.

davepagurek:
Resolves https://github.com/processing/p5.js-website/issues/518

It looks like we add a stroke to text to make it bolder. There were a few cases where we would update the fill color of text but not the stroke color, leading to a slight different-colored halo around text. It was happening in links before, see below:

Before:
image

After:
image

That said, it looks like this text stroke may be the cause of https://github.com/processing/p5.js-website/issues/473. I'm not sure how standard this property is, and whether it's a good idea to be using it across the board on body text like this. Possibly a sign that we should just be using a thicker font instead?

### DIFF
--- a/styles/global.scss
+++ b/styles/global.scss
@@ -186,7 +186,7 @@ h4 {
 body {
   background-color: var(--bg-color);
   color: var(--type-color);
-  -webkit-text-stroke-color: var(--type-color);
+  -webkit-text-stroke-color: currentcolor;
   @extend .text-body;
   font-weight: 400;
 }
@@ -315,7 +315,6 @@ pre.code-box,
 .sidebar {
   background-color: var(--sidebar-bg-color);
   color: var(--sidebar-type-color);
-  -webkit-text-stroke-color: var(--sidebar-type-color);
 }
 
 /** LAYOUT **/


Qianqianye:
Thanks @davepagurek. I'm merging this solution for now, and we can have schedule a chat with the web designer with all the [issues](https://github.com/processing/p5.js-website/issues?q=is%3Aopen+is%3Aissue+label%3A%22Web+Desigb%22) and [PRs](https://github.com/processing/p5.js-website/pulls?q=is%3Apr+is%3Aopen+label%3A%22Web+Desigb%22) labelled `Web Design` soon.

PR #560, opened 2024-09-20
Fix Empty examples are shown on several reference pages

shibomb:
resolves #559

## Changes
Corrected separator strings in some reference page examples.

PR #561, opened 2024-09-21
fix: Added process to remove last empty row for examples in function references.

### LINKED ISSUE #559: Empty examples are shown on several reference pages.

Reference

None

### Web browser and version

_No response_

### Actual Behavior

Empty examples are shown on several reference pages.
e.g. https://p5js.org/reference/p5/background/

スクリーンショット 2024-09-20 20 59 03

### Expected Behavior

Empty examples are not displayed.

### Steps to reproduce

1. Open https://p5js.org/reference/p5/background/
2. Scroll to the end of the example.

I will create PR

shibomb:
resolves #559
see detail: https://github.com/processing/p5.js-website/pull/560#issuecomment-2365070020

### DIFF
--- a/src/scripts/builders/reference.ts
+++ b/src/scripts/builders/reference.ts
@@ -152,7 +152,7 @@ const correctRelativeLinksToExampleAssets = (
     return content;
   }
   return Array.isArray(content)
-    ? content.map((ex) => ex.replaceAll("assets/", "/assets/"))
+    ? content.map((ex) => ex.replaceAll("assets/", "/assets/").replaceAll(/\n$/g, ""))
     : content.replaceAll("assets/", "/assets/");
 };
 


src/scripts/builders/reference.ts
```diff
@@ -152,7 +152,7 @@ const correctRelativeLinksToExampleAssets = (
     return content;
   }
   return Array.isArray(content)
-    ? content.map((ex) => ex.replaceAll("assets/", "/assets/"))
+    ? content.map((ex) => ex.replaceAll("assets/", "/assets/").replaceAll(/\n$/g, ""))
```
davepagurek, line 155: Thanks for adding this! Two thoughts:
- Do you think we should also add it on the line below in the case when the content is just a string?
- To make things easier for other contributors coming into this in the future, maybe we should make a new function like `removeTrailingNewlines()` to do this functionality and then call it in places where we call `correctRelativeLinksToExampleAssets()`? It might be unexpected, based on the name of the function, that it would include this functionality. Alternatively, maybe we can rename this to something more general, like `cleanUpExampleContents`?

shibomb: As you pointed out, you are right, the createRelativeLinksToExampleAssets function is indeed for modifying the links.
This was not the appropriate part of this response...

I like the idea of either removeTrailingNewlines() or cleanUpExampleContents().

( Or, I still think the policy should be to fix it in the p5.js source code... )

davepagurek: I think fixing it in the p5 repo also works, although we'd maybe want to add a check in CI that there aren't trailing newlines, because I can imagine it being pretty easy to miss in code review otherwise.

But in the mean time, renaming this function or putting this change in another function that's called alongside this one is also fine by me! Thanks for finding the cause and for making these changes so far 🙂

shibomb: Thank you too.
I am happy to help improve p5.js!

limzykenneth:
@shibomb Just one administrative thing before I merge this, does this supercede #560?

shibomb:
> @shibomb Just one administrative thing before I merge this, does this supercede #560?

Thanks for the review!
If you merge this #561 and run “npm run build:reference” next time, each mdx file will be similar to #560.
So #560 is not necessary. 
I will withdraw the pull request ;-)

limzykenneth:
Sorry for the delay. Looks good. Thanks!

Issue #564, opened 2024-09-22
The parameter type is not displayed.


shibomb:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

None

### Actual Behavior

The parameter type is not displayed.
In mdx files with simple params, the parameter types are not displayed.
e.g. circle()

But an mdx file with overloads will show the parameter types.
e.g. line()

![スクリーンショット 2024-09-23 4 41 53](https://github.com/user-attachments/assets/8783dd2d-0321-42a9-8e7e-02b6c6217248)

### Expected Behavior

The parameter type is always displayed, with or without overloading.

### Steps to reproduce

open https://p5js.org/reference/p5/circle/ and https://p5js.org/reference/p5/line/ .
compare Parameters section.

### Would you like to work on the issue?

I will create PR

davepagurek:
Resolved by #565

PR #565, opened 2024-09-22
Fix: Show param type and return in reference page.

### LINKED ISSUE #562: The “Returns” section does not appear on the reference page for functions and methods that return values.

Reference

None

### Web browser and version

_No response_

### Actual Behavior

The “Returns” section does not appear on the reference page for functions and methods that return values.

スクリーンショット 2024-09-21 20 42 40

### Expected Behavior

The “Returns” section does appear on the reference page for functions and methods that return values.

### Steps to reproduce

1. open https://p5js.org/reference/p5/cos/
2. scroll down to under Parameters.

e.g. function:
https://p5js.org/reference/p5/cos/
(old version: https://archive.p5js.org/reference/#/p5/cos )

e.g. method: 
https://p5js.org/reference/p5.Vector/copy/
(old version: https://archive.p5js.org/reference/#/p5.Vector/copy )

I will create PR

shibomb:
Resolves #562 #564 
Sorry for multiple issues.

### Changes:
- for #564 : param type shows with simple param field.
- for #562 : show return info. and add UI label (en.yaml) for translations.


src/layouts/ReferenceItemLayout.astro
```diff
@@ -169,6 +171,23 @@ const seenParams: Record<string, true> = {};
           </>
         )
       }
+      {
+        (entry.data.return) && (
+          <>
+            <div class="mb-xl">
+              <h2 class="mb-md text-h3">{t("Return")}</h2>
```
davepagurek: Looks like on the old site it was "Returns" with an s, maybe we can do the same here too?

image


```suggestion
              <h2 class="mb-md text-h3">{t("Returns")}</h2>
```

davepagurek (commented):
Thanks, looks good! I just noticed one difference from before, but otherwise good to go.

src/layouts/ReferenceItemLayout.astro
shibomb: @davepagurek 
Thanks for reply.

This is a suggestion for improvement.

Indeed, the old version was with "s".
I assume this “s” means plural, but can the returned value be plural?

- It is certainly possible to have multiple meanings by using arrays or Json, but I think that should be explained in the Description.

- As a property of the .mdx file is “return:”.

- Also, for example, there is no “s” in processing.org.
  (I haven't checked all the pages... For example, https://processing.org/reference/cos_.html .)

Please consider the above!

If you need the “s”, I will modify the necessary part :-)

davepagurek: Returns with an s is I think more common, e.g. [Tone.js](https://tonejs.github.io/docs/15.0.4/classes/Context.html#createWaveShaper), [React](https://react.dev/reference/react/useCallback#returns), [matter.js](https://brm.io/matter-js/docs/classes/Constraint.html), and [lodash](https://lodash.com/docs/4.17.15#now). I think the reason for it isn't to consider "return" as a plural noun, but rather as a verb, e.g. "this function returns a number", or just "returns: number" for short in this case. In this case the s just means present tense, "it returns" as opposed to "it returned" in the past.

shibomb: I understand.
I will correct it!

shibomb: @davepagurek 
I have updated the branch.
Please re-review.
Thank you.

davepagurek (approved):
Thanks, looks great!

PR #566, opened 2024-09-22
restore deleted rita library

dhowe:
restore deleted rita library (not sure why this was deleted in the first place)  
please check @outofambit @davepagurek


src/content/libraries/en/rita.js.yaml
```diff
@@ -0,0 +1,13 @@
...
+category: language
+sourceUrl: https://github.com/dhowe/ritajs
+websiteUrl: https://rednoise.org/rita
+featuredImage: "../images/rita.jpg"
+author:
+  name: Daniel C Howe
+  url: https://rednoise.org/daniel
+featuredImageAlt: rita.js logo
+npm: rita
+npmFilePath: "https://unpkg.com/rita@3.1.3/dist/rita.min.js"
```
davepagurek: I think in this case you'll just need the bit after the version, similar to https://github.com/processing/p5.js-website/blob/df3d65019dafc24e54e8cb429cc9e38b1618c230/src/content/libraries/es/p5.xr.yaml#L12
```suggestion
npmFilePath: "dist/rita.min.js"
```

davepagurek (commented):
Thanks for adding this! It wasn't that things got deleted, but that we needed new data from library authors, so we just asked people to submit things again. Thanks for submitting rita.js!

dhowe:
Sure, but I at least was _never_ asked. And I see other libraries missing and wonder if those authors were queried

Of course its not your fault, but I have to say its very frustrating to spend so much time on such things and then find them simply removed without so much as a heads-up...

davepagurek:
We tried to send an email to everyone who submitted a library before, but I feel like there's a good chance that ends up in spam if the Google gods aren't feeling it on a given day. Sorry it never got to you!

dhowe:
Again, not trying to be rude, but there aren't so many people involved here - it seems a bit more of an effort could have been made (library authors are relatively easy to find on the web)

davepagurek:
I hear you. It was a conscious decision to not send stuff to authors beyond that initial email before getting the site out to save on organizational time with all the other moving parts (I'm sure you've noticed other bugs on the new site that we're still in the process of fixing), so we take responsibility for that. Sorry it's affected you!

Issue #567, opened 2024-09-23
The order of the jump-to menus on the tutorial page is alphabetical.


shibomb:
### Most appropriate sections of the p5.js website?

Tutorials

### What is your operating system?

None

### Actual Behavior

The order of the jump-to menus on the tutorial page is alphabetical.

### Expected Behavior

The order of the jump-to menus on the tutorial page is the same as the order in which they are listed on https://p5js.org/tutorials/.

### Steps to reproduce

open https://p5js.org/tutorials/setting-up-your-environment/
see jumpTo menu on left side bar.

### Would you like to work on the issue?

I would make a pull request for improvement ideas.

davepagurek:
@shibomb do you think any more changes are needed here, or are we free to close this issue now?

shibomb:
@davepagurek 
If my implementation is OK, the problem is solved.
We have no problem with closing it.
Thank you!

davepagurek:
It looks good to me 🙂 Thanks!

PR #568, opened 2024-09-23
fix jumpTo menu order in tutorial pages.

shibomb:
ideas for  #567

This is an improvement idea that we tried in the Japanese version.
https://p5js-ja.pages.dev/tutorials/setting-up-your-environment/

However, this implementation may not be the best due to my lack of study:.
- I am changing constants to variables.
- TypeScript type safety is not observed.

You may want to consider another better implementation.

I hope this helps you.


davepagurek (approved):
Good catch, thanks!

PR #574, opened 2024-09-25
Fix of Educational Site

sebastian-meier:
The content shows up twice. Remove the Content from the site (as its already included in the BaseLayout.

See also current version here: https://p5js.org/education-resources/


src/pages/education-resources/index.astro
```diff
@@ -39,5 +39,4 @@ setJumpToState({
   variant="item"
   topic="tutorials"
 >
-  <Content />
```
davepagurek, line 43: Thanks for the fix! Since we're no longer using this, I think we can delete these lines too, which I believe are only used for creating the `Content` component used here:
https://github.com/processing/p5.js-website/blob/6a83470e7cf64bc30d845f19c8bf01089d559188/src/pages/education-resources/index.astro#L10-L13

sebastian-meier: I removed the last line. The other two are actually requried.

PR #575, opened 2024-09-25
include the p5.colorGenerator library

alexandru-postolache:
Hello, I created a library to generate color palettes for p5.js, and I would like to include it on the website.


davepagurek (approved):
Looks great, thanks!

PR #578, opened 2024-09-26
Update reference for v11, fix locked version on p5 update

davepagurek:
**Note: do not merge this until https://cdnjs.cloudflare.com/ajax/libs/p5.js/1.11.0/p5.min.js is live!** It updates src/globals/p5-version.ts, which will cause the reference to use the new built library.

Previously, I had locked the version number in the reference building script to v1.10.0 so we could rebuild it whenever without getting newer, unreleased contents from main. Now that we've done a v1.11.0 release, I realize we already have a p5 version constant in our codebase, so when we updated, the reference was still being built from the previous release.

I've updated the reference builder script to now use the p5 version that is outputted by `npm run build:p5-version`, and I've rebuilt the reference with the new content.


davepagurek:
Looks like it's a cdnjs issue: https://github.com/cdnjs/cdnjs/issues/14261

davepagurek:
If we want to use another CDN in the mean time, https://cdn.jsdelivr.net/npm/p5@1.11.0/lib/p5.min.js is working.

Edit: just put up https://github.com/processing/p5.js-website/pull/579 in case we want to release using an alternative CDN

Qianqianye:
Thanks @davepagurek!

PR #579, opened 2024-09-28
Use jsdelivr as CDN

davepagurek:
This branches off of https://github.com/processing/p5.js-website/pull/578 and uses a different CDN provider that has the new p5 version. Feel free to close this if we want to wait.


Qianqianye:
We've been thinking about switching to jsdelivr, so let's just try it. We can always switch back to cdnjs if needed when the [cdnjs outage issue](https://github.com/cdnjs/cdnjs/issues/14261) is resolved.

CC @raclim @limzykenneth

raclim:
Thanks for the heads up!

Issue #585, opened 2024-10-07
The JumpTo category name are appear in lowercase in the left sidebar of the example child page.


shibomb:
### Most appropriate sections of the p5.js website?

Examples

### What is your operating system?

None

### Actual Behavior

The JumpTo category name are appear in lowercase in the left sidebar of the example child page.
(It will not be translated in any language other than English.)

### Expected Behavior

Display the same as it appears on https://p5js.org/examples

### Steps to reproduce

open https://p5js.org/examples/shapes-and-color-shape-primitives/
see JumpTo link's category in left side bar.

compare with https://p5js.org/examples/

### Would you like to work on the issue?

I would create a pull request for suggestions.

limzykenneth:
Fixed with #586

PR #586, opened 2024-10-07
FIX the category name for the JumpTo menu in the example child pages.

shibomb:
Resloves: #585

This solution is a suggestion.
It appears to be working well in the Japanese version.
https://p5js-ja.pages.dev/en/examples/shapes-and-color-shape-primitives/


limzykenneth:
Looks good. Thanks!

PR #588, opened 2024-10-08
Fix: The “Related Reference” section title of the example pages for i18n

shibomb:
The “Related Reference” section title of the example pages has been revised to include internationalized words as well as other pages.

Changes:
"Related Reference" to "Related References"


limzykenneth:
Looks good. Thanks!

PR #589, opened 2024-10-08
Fix example category link in the jumpTo on the example detail page.

shibomb:
(Relation #585, #586)

Example Category link broken after merge #586.

Details:
Before merge #586, there was no problem in English. 
But after merged #586, the problem was discovered because #ID and id attribute were created with different rules.

- Category's id attribute in parent page was created with "name.toLowerCase()".
https://github.com/processing/p5.js-website/blob/d24bb80293c026f1bb688f3b4c2acdcf3a358bc6/src/layouts/ExamplesLayout.astro#L45

- jumpTo Category's #ID was created with "category".

Changes:
- jumpTo Category's #ID is created with "name" and ".toLowerCase()".

### DIFF
--- a/src/pages/_utils.ts
+++ b/src/pages/_utils.ts
@@ -367,7 +367,10 @@ export const generateJumpToState = async (
     const categoryLinks = [] as JumpToLink[];
     categoryLinks.push({
       label: getCategoryLabel(category),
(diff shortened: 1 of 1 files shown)


Qianqianye:
Thanks @shibomb!

PR #591, opened 2024-10-09
Resolves #584

### LINKED ISSUE #584: Cannot translate "Listing Data with Arrays" in JumpTo Sidebar on Examples page.

Examples

None

### Web browser and version

_No response_

### Actual Behavior

Cannot translate the “Listing Data with Arrays” in the JumpTo sidebar on the example page.
(This occurred when I was translating the Japanese version.)

### Expected Behavior

It is translated correctly.

This name is generated from the folder name “08_Listing_Data_with_Arrays”.
Based on the other folder names in content/examples/en/, it may be correct to change the “with” in this folder name to “With”.

Then "Listing Data With Arrays"("W" is Upper Case.) will match the word defined in ui/en.yaml :-)

スクリーンショット 2024-10-07 22 02 15

### Steps to reproduce

This does not occur in the current source code.
However, it will occur in the future when you translate src/content/ui/es.yaml etc. in sync with en.yaml.

It is difficult for me to change folder names from lower case to upper case.

rthakur2712:
resolves #584 
Renamed all the directories containing the "Listing Data with Arrays".


davepagurek:
Were you trying to change the capitalization of a file? It looks like it hasn't shown up in got if you check the Files Changes tab of this PR because git generally treats files as being case insensitive.

Maybe if you change it back, then try some of the suggestions here https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git it'll work?

rthakur2712:
Yes got it thanks for the information. Since git ignores change in case of a directory name i thought to change the directory name first to something random and then to the desired name and I think it should work properly. Please verify if the desired changes are reflecting.

ksen0:
Hi @rthakur2712 , sorry I didn't see this until now! Could you check out [the comment](https://github.com/processing/p5.js-website/issues/584#issuecomment-2627067907) on the issue? I can see this is a challenge but if there's a way to address it without changing directory capitalization I think that could have fewer possible side effects! Thanks so much for your patience.

PR #605, opened 2024-10-17
Add script to preview a custom branch

davepagurek:
This adds a script, `npm run custom:dev https://github.com/yourUsername/p5.js.git#yourBranch`, that builds the reference off of your fork of p5 and starts a dev server (and also `npm run custom:cleanup` to remove the reference changes afterwards.)

It also makes the reference schema a little more forgiving, making more things optional with fallbacks, as these builds are less likely to be entirely complete.


davepagurek:
@limzykenneth This changes how the p5 reference is built, now just `npm install`ing and calling `npm run docs` directly. This is maybe a little slower for 1.x when we could just run yuidoc directly, but it means it builds the docs correctly for both 1.x and 2.x.

I'm hoping this will make it easier for contributors to be able to preview their docs changes.

PR #606, opened 2024-10-18
fix: Skip escaping if already escaped in the code tag.

### LINKED ISSUE #587: Reference shows html entity &amp; rather than ampersand
https://p5js.org/reference/p5/if/

image

Reference

Mac OS

### Web browser and version

Firefox 131.0 (64-bit)

### Actual Behavior

HTML entity code is shown

### Expected Behavior

actual character should be shown

### Steps to reproduce

visit and look

no

shibomb:
Resolves #587 

This is a proposal.
It seems to work well on the following page in my local env.
/reference/p5/createInput/
/reference/p5/Boolean/
/reference/p5/Number/

## changes:
Check if the content is already escaped by looking for common HTML entities in code tag.
If already escaped, skip escaping.

### DIFF
--- a/src/pages/_utils.ts
+++ b/src/pages/_utils.ts
@@ -239,6 +239,18 @@ export const escapeCodeTagsContent = (htmlString: string): string => {
 
     // Get the current text and HTML inside the <code> tag
     const currentHtml = $(this).html() ?? "";
+
+    // Check if the content is already escaped by looking for common HTML entities
+    if (
+      currentHtml.includes("&lt;") ||
+      currentHtml.includes("&gt;") ||
+      currentHtml.includes("&amp;") ||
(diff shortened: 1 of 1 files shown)


davepagurek:
This is a good fix for right now while we have stuff written in an inconsistent format. We should also address that at the source (details in https://github.com/processing/p5.js-website/issues/587#issuecomment-2422407707) but this will help us not have broken content in the mean time!

davepagurek:
One note is that it means we can't easily write code in p5 explaining how HTML escaping works because if we try to explicitly write `&lt;` as I've done here, it would render as `<` in our docs. I think that's fine since we probably don't need to do that in p5, but once the source is addressed, the plan will be to remove this check again.

PR #615, opened 2024-10-20
Rework compression for faster build

limzykenneth:
`astro-compress` leaves too much possible performance on the table by not processing images concurrently. This version tries to do this as much as possible, one possible bottle neck is IO. Possible way to optimize this is through utilizing streams but will require temporary write location.

Better build console output would be needed as well.


limzykenneth:
Build time decreased from last deployed `1180.74s` to `788.50s` in CI. About `461.65s` on my computer. GitHub Actions likely don't provide as much processing power but it should make things a bit more speedy unless we ever want to use GitHub Actions build artifacts.

limzykenneth:
@davepagurek If you have the time, do have a try to build from this PR and see if there are any problems. The overall build time should be about half as before now. No worries if you are busy with other stuff though as this is not urgent.

ksen0:
I've checked it locally, works!

Issue #627, opened 2024-11-03
Embedded p5.js is not displayed in the local development environment.


shibomb:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

None

### Actual Behavior

Embedded p5.js is not displayed in the local development environment.
<img width="1402" alt="スクリーンショット 2024-11-03 12 57 33" src="https://github.com/user-attachments/assets/7ff8d533-447a-414d-9c65-c58698edb42b">

console displays an error:
```
arc/:35 [astro-island] Error hydrating /src/components/CodeEmbed/index.jsx ReferenceError: process is not defined
    at globals.ts:17:3
```

### Expected Behavior

Embedded p5.js is displayed in the local development environment.

### Steps to reproduce

1. in local pc, run "npm run dev" with commit: https://github.com/processing/p5.js-website/pull/622
2. see examples of http://localhost:4321/reference/p5/arc/ 

### Would you like to work on the issue?

I will create PR

PR #628, opened 2024-11-03
fix: "process" not found on local dev environment

shibomb:
Solves #627

changes:
It also determines if the process object is undefined.


davepagurek:
I think we still need to read the env vars in order for testing custom p5 builds to work, so I think we can't just fall back on the CDN here. I'm going to look into an alternate way to pass that data in that works with hydrated components too.

shibomb:
@davepagurek Thank you for your response!
It works fine in my local environment with #629.
I'll close this pull request.

Issue #597, opened 2024-10-15
"Move to Top" Button


solanki505:
### Increasing Access

this will reduce the time to scroll back at top

### Most appropriate sub-area of p5.js?

Home

### Feature request details

makes the site more user friendly

rishabdev2997:
Hi @limzykenneth, @lmccart, @kjhollen

I'd like to discuss the implementation of the "Move to Top" button for the site. I’ve created a component for it and used SCSS for styling, but before submitting a new PR, I want to gather feedback to ensure it aligns with our overall design and approach. Here are a few points for discussion:

Feature Necessity — Do we think this button would add value to the user experience? If not, we can close the issue.
Design and Placement — The button is currently positioned in the bottom right corner. Should we consider any adjustments to its design or placement?
Consistency with Site Styling — I’ve used SCSS for styling. Please let me know if this approach aligns with the current best practices we follow.
Looking forward to your feedback and thoughts before moving forward.

Thanks!

limzykenneth:
Just in terms of the styling method (not commenting on the necessity and design aspects at the moment), many of the styling that we use on the website uses TailwindJS and not necessarily SCSS. Tailwind can help reduce the overall code duplication in CSS so I would prefer to use it where possible, so this should be looked into.

rishabdev2997:
Hi @limzykenneth, thank you for your feedback! I understand your point about preferring TailwindCSS for styling consistency across the project. Before I go ahead and refactor the implementation to use Tailwind, I wanted to clarify if it would be better to first discuss the necessity and design of this feature in more detail. Please let me know how you'd like to proceed, and I'm happy to make any adjustments based on the project’s needs. Thanks for guiding me through this!

limzykenneth:
Yes, there need to be some discussion and research into the necessity and design. For example (not the only thing to determine), should go to top button be animating to the top or just jumps to it? Should it use a `#top` id or use JS to set scroll position. There are many ways to implement this and we need to determine which fits with the overall design of the website first.

rishabdev2997:
Hi @limzykenneth thank you for clarifying! I understand now that we need to discuss the specific design and implementation details before proceeding. For example, as you mentioned, whether the button should animate smoothly to the top or just jump, and whether to use an #id or JavaScript for scrolling behavior. I agree it’s important to align this with the overall design approach of the website.
[...]

PR #631, opened 2024-11-08
[Feat] :Adding a move to top icon

rishabdev2997:
### [Feat] : Adding a move to top icon
#597 - Added the feature.
### Feature details-
**This feature reduces the time and effort needed to manually scroll up, making navigation smoother.**
### How to see the change
- **Scroll Button Location:** The "Move to Top" button is positioned at the bottom-right corner of the homepage.
- **Visibility Condition:** The button becomes visible only when the user has scrolled down from the top of the page, enhancing usability by not cluttering the view initially.
- **Functionality:** Clicking on the button will instantly scroll the user back to the top of the page.

### Here is the screenshot

https://github.com/user-attachments/assets/55ca8a2b-671f-4811-85aa-e66c9a18d001

### DIFF
--- a/src/components/MoveToTopButton/index.astro
+++ b/src/components/MoveToTopButton/index.astro
@@ -1,0 +1,38 @@
+---
+const { class: customClass = "" } = Astro.props;
+---
+
+<a
+  id="moveToTop"
+  class={`move-to-top hidden outline outline-1 outline-type-color py-2 px-4 flex flex-nowrap w-fit items-center justify-center rounded-full hover:bg-sidebar-type-color hover:text-bg-color hover:no-underline ${customClass}`}
+  href="#"
+  title="Go to top"
+>
+  <span>↑</span>
+</a>
+
+<script>
+  document.addEventListener("DOMContentLoaded", () => {
+    const moveToTopButton = document.getElementById("moveToTop");
+
+    // Function to toggle visibility based on scroll position
+    const toggleVisibility = () => {
+      if (window.scrollY > 100) {
+        moveToTopButton.classList.remove("hidden");
+        moveToTopButton.classList.add("visible");
+      } else {
+        moveToTopButton.classList.add("hidden");
+        moveToTopButton.classList.remove("visible");
+      }
+    };
+
+    // Attach scroll event listener
+    window.addEventListener("scroll", toggleVisibility);
+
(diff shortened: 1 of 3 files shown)


rishabdev2997:
Memory updated
Here’s a message you can send to Limzy:

Hi @limzykenneth,

I hope you're doing well! I’ve made a PR for the "Move to Top" button component. Previously, I had submitted a version with inline styles, which is generally considered bad practice. This time, I've refactored it to avoid inline styles and implemented the button via a component-based approach.

Could you please review the PR and let me know if any changes are required or if you have any feedback? Once reviewed, please feel free to merge it.

Looking forward to your thoughts!

Thanks,
@rishabdev2997

rishabdev2997:
Hi @limzykenneth,

I hope you're doing well! I’ve made a PR for the "Move to Top" button component. Previously, I had submitted a version with inline styles, which is generally considered bad practice. This time, I've refactored it to avoid inline styles and implemented the button via a component-based approach.

Could you please review the PR and let me know if any changes are required or if you have any feedback? Once reviewed, please feel free to merge it.

Looking forward to your thoughts!

Thanks,
@rishabdev2997

limzykenneth:
Thanks @rishabdev2997. As mentioned in your previous PR, there really should be some discussions in the issue to determine the implementation first. This is to serve three purposes:

1. Discuss and determine if the feature is really needed or not.
2. If decided that it is needed, what the proposed design and implementation should be.
3. Understanding the existing approach and using a consistent implementation approach.

The implementation you have here does not match the general approach we take to adding and styling elements on the website. I'm hesitating to directly mention the general direction in terms of the approach here because I really need a more detail discussion to be had in the issue first before a PR should be filed.

rishabdev2997:
Hi @limzykenneth ,

Thank you for the feedback, and I understand your concerns. I apologize for jumping directly into the PR without a detailed discussion in the issue thread. I wanted to address the feedback from the previous PR about avoiding inline styles, so I went ahead with a component-based approach this time.

I agree that it’s essential to have a consistent implementation and design approach across the website. I’ll take a step back and initiate a discussion in the relevant issue to align on the need for this feature and determine the appropriate design and implementation method.

Looking forward to collaborating on this and making sure it fits well with the overall project standards.

Thanks again for your guidance!

Best,
@rishabdev2997

PR #632, opened 2024-11-09
added padding in examples header

### LINKED ISSUE #573: Add `Adequate Spacing` Between Text and Images for Improved Visibility in `Example Page`
### Increasing Access

Improve Design and Readability

Reference

### Feature request details

There is currently insufficient spacing between the text and images in the Shapes And Color and Animation And Variables sections. The proximity of the text to the images makes it harder for users to distinguish between the two, affecting readability and overall user experience.

**Actual Behavior:** 

image

**Expected Behavior:** 

image

**Proposed Solution:**

- Add a margin to the images and padding to the text elements.
- Ensure consistent spacing across all sections for uniformity.
- Test the interface for responsiveness to make sure the changes apply well across different screen sizes.

mohankumarhr:
Fixes #573 
A padding is added for the headings of the Example section to give gap between heading and images
Screenshot (23)

### DIFF
--- a/src/layouts/CommunityLayout.astro
+++ b/src/layouts/CommunityLayout.astro
@@ -45,7 +45,7 @@ setJumpToState({
 
 <BaseLayout title="Community">
   <section>
-    <h2>{t("Sketches")}<a id="sketches"></a></h2>
+    <h2 class="mb-gutter-md mt-0">{t("Sketches")}<a id="sketches"></a></h2>
     <ul class="content-grid-simple">
       {
         sketches.map((sk, i) => (
@@ -74,7 +74,7 @@ setJumpToState({
   </section>
 
   <section>
-    <h2>{t("Libraries")}<a id="libraries"></a></h2>
+    <h2 class="mb-gutter-md mt-0">{t("Libraries")}<a id="libraries"></a></h2>
     <ul class="content-grid-simple">
       {
         libraries.map((lib) => (
@@ -94,7 +94,7 @@ setJumpToState({
   </section>
 
   <section>
-    <h2>{t("Events")}<a id="events"></a></h2>
+    <h2 class="mb-gutter-md mt-0">{t("Events")}<a id="events"></a></h2>
     <ul class="content-grid-simple">
       {
         events.map((event, i) => (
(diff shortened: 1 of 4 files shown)


davepagurek:
Hi! This will affect *all* h2s, probably breaking the layout in tutorials that include it. The tutorials page has a similar look and addresses it by adding the following tailwind class to their h2s:
https://github.com/processing/p5.js-website/blob/753bb06aba1221c5a60cb6158fa806d66e8a71eb/src/layouts/TutorialsLayout.astro#L53

Maybe we can take the same approach here instead?

mohankumarhr:
> Hi! This will affect _all_ h2s, probably breaking the layout in tutorials that include it. The tutorials page has a similar look and addresses it by adding the following tailwind class to their h2s:
> 
> https://github.com/processing/p5.js-website/blob/753bb06aba1221c5a60cb6158fa806d66e8a71eb/src/layouts/TutorialsLayout.astro#L53
> 
> Maybe we can take the same approach here instead?

Thank you for the suggestion @davepagurek. I have removed the previous update and added the Tailwind class

Nwakaego-Ego:
Hi @mohankumarhr, thank you for your work on this so far. The spacing values used in the PR appear to be correct: margin-top: 0 and margin-bottom: var(--gutter-md). However, when I tested the PR, the mt-0 mb-gutter-md classes were present in both the source and the rendered h2, but DevTools still showed:
margin-top: 40px;
margin-bottom: 0;
It looks like these classes are being overridden by the existing section-heading rule in global.scss:
h2,
.text-h2 {
  margin-top: var(--gutter-md);
  margin-bottom: 0px;
}

Because of this, the spacing issue is still visible on the Examples page when the PR is applied to the current v1 branch.
To compare the results, I tested the values directly on the matching section h2 rule. This created the expected spacing on the Examples pages in both v1 and main. The Tutorials directory also looked fine at a 445px viewport. On the individual tutorial page I checked, the article headings kept their existing 32px top margin because they are inside .rendered-markdown, not inside a <section>.
I also noticed the same spacing pattern on other pages, so the changes outside ExamplesLayout.astro may be relevant. However, since a shared styling change could affect more pages, it would be helpful to explore which  approach would give the intended spacing without changing unrelated headings.
Are you available to update the PR from the current v1 branch and explore a fix that works with the current CSS setup? Before and after screenshots at desktop and mobile widths would also help us confirm the result. The same issue is present on main, although this PR currently targets v1. Thank you.

PR #694, opened 2025-01-30
Fix html minifier removing relevant spaces in html body

### LINKED ISSUE #693: Incorrect spacing around all inline code snippets and links

Other (specify if possible)

Windows

### Web browser and version

Chrome: 132.0.6834.111 (Official Build) (64-bit) (cohort: Stable), Edge: 132.0.2957.127 (Official build) (64 bits), Firefox: 121.0.1

### Actual Behavior

For both inline code snippets (highlighted in grey) and links (in red), leading and trailing spaces are removed, resulting in text that's hard to read. This seems to be true throughout the website, in both the reference and non-reference pages. An example is below.

Image

Source: [https://p5js.org/reference/p5/arc/](https://p5js.org/reference/p5/arc/) 

### Expected Behavior

I expect inline code snippets and links to be separated from surrounding text by spaces. This behavior can be seen, for example, in the archived version of the old p5.js site. An example is below.

Image

Source: [https://archive.p5js.org/reference/#/p5/arc](https://archive.p5js.org/reference/#/p5/arc)

### Steps to reproduce

The behavior is present in all three of the browsers I've checked, so I suspect it can be reproduced by opening any page on the website in any browser, probably in any environment.

If someone else can work on this, that'd be great!

limzykenneth:
Fixes #693

### DIFF
--- a/src/scripts/fast-compress.ts
+++ b/src/scripts/fast-compress.ts
@@ -90,7 +90,7 @@ Writing    ${writingBar} ${Math.floor(writingProgress * 100)}% (${total - writin
 
 						try{
 							const data = await minify(text, {
-								collapseWhitespaces: "all",
+								collapseWhitespaces: "smart",
 								removeComments: true,
 								sortAttributes: true
 							});


davepagurek (approved):
Thanks for looking into this!

PR #700, opened 2025-02-13
Submission for Turtle Graphics library

quarks:


davepagurek:
Thanks @quarks! @ksen0 and @limzykenneth, do you know if the netlify bot failures are blocking merging a PR like this? Other tests are all good to go.

limzykenneth:
I don't think it is blocking but it probably shouldn't be running I think. @ksen0 Can you turn off builds for netlify? We can setup the deploy from Github Actions instead of having netlify pull from GitHub.

PR #706, opened 2025-02-20
fixed typo: frameCount is a variable not a function

axwnet:
The text of the reference on the [`draw()`](https://p5js.org/reference/p5/draw/) page clearly states that `frameCount` is a system variable. Therefore `frameCount()` is not correct.


perminder-17 (approved):
Thankyou for catching this!

Issue #709, opened 2025-02-27
AnnotatedCode component scrolls in Chrome


davepagurek:
### Most appropriate sections of the p5.js website?

Tutorials

### What is your operating system?

Mac OS

### Web browser and version

Chrome

### Actual Behavior

When the right side of a section of `AnnotatedCode` is smaller than the left side with the code, the code scrolls. This appears in Chrome but not Firefox, and can be seen in Intro to Shaders in the section on varying variables: https://p5js.org/tutorials/intro-to-shaders/

<img width="1358" alt="Image" src="https://github.com/user-attachments/assets/b71e5199-4e0a-4cd6-aeb1-fa6f441b1df6" />

### Expected Behavior

The code cell should expand to fit the code and not scroll

### Would you like to work on the issue?

Someone else can take this on if you're interested!

webermayank:
Hi @davepagurek 
I am interested in taking this issue, and would like to work on this.

davepagurek:
Thanks @webermayank, I'll assign this to you!

PR #711, opened 2025-03-02
fix : #709 Scroll removed and code cell now expands to fit code

webermayank:
Fixes issue #709
Here are the screenshots of before and after the fix
#### Before :
Screenshot 2025-03-02 131321

#### After :
Screenshot 2025-03-02 131109

@davepagurek
Let me know if there are any other changes require, i would love to address them.


davepagurek:
Thanks for looking into this so far! It looks like now there's a gap between the table cells in the left column where before they connected. Ideally the left column still visually looks like one unbroken code block. Is there a way to still maintain that, but also not have it scroll?

webermayank:
@davepagurek 

yes it is possible to maintain a code cells like a single code block and not have gaps between them
#### screenshot after the fix
Screenshot 2025-03-03 090908

now the code cells doesn't have scrolls and gaps between them 

Tell me if anything seems off, i would love to fix it.

davepagurek:
I think this looks good, thanks for making the update!

webermayank:
Its all my pleasure.
Looking forward to be more helpful

PR #719, opened 2025-03-12
Clarify that `calculateBoundingBox` is aligned to the axes.

Invariance-NaN:
The current docs for `calculateBoundingBox` are not quite correct, since the "smallest rectangular prism" containing the geometry need not be aligned to the axes.

### DIFF
--- a/src/content/reference/en/p5.Geometry/calculateBoundingBox.mdx
+++ b/src/content/reference/en/p5.Geometry/calculateBoundingBox.mdx
@@ -4,10 +4,10 @@ module: Shape
 submodule: 3D Primitives
 file: src/webgl/p5.Geometry.js
 description: >
-  <p>Calculates the position and size of the smallest box that contains the
+  <p>Calculates the position and size of the smallest axes-aligned box that contains the
   geometry.</p>
 
-  <p>A bounding box is the smallest rectangular prism that contains the entire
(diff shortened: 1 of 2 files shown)


ksen0:
Thanks for raising this @Invariance-NaN !

The reference is automatically built from the `p5.js` core source code. To change this part of the reference, this file should be changed: https://github.com/processing/p5.js/blob/db7747174b32ab1851547f0212e5b8e24d2064cb/src/webgl/p5.Geometry.js#L683 As suggested in the contributor guidelines, in the future / in general, please make an issue first, and explain your proposal there, before working on a solution - so that the solution can first be discussed.

I'll close this PR lease feel free to open an issue/PR in that repository!

In terms of the suggestion itself, I think pointing out that the outcome needs, e.g., translation (in the example `translate(bbox.offset.x, bbox.offset.y, bbox.offset.z);`) would be helpful, however I think the first line of the reference should remain as beginner-friendly as possible. To strike a balance, I'd suggest adding a 1 or 2 sentence paragraph, after "size (length) and offset (center).", that uses the axis-aligned bounding box terminology, and also explains it. I think it's important that if we use any technical terms, we also explain them in the reference, however I might be wrong in my understanding of the intent here, or maybe others have a better idea.

If you do make an issue/PR in the `p5.js` repository to make this suggestion, or an updated version, please feel free to tag me or @perminder-17 and one of us will be happy to review!

PR #728, opened 2025-03-15
fixed yuidoc path for build reference.ts

webermayank:
Previously if the user has space in their path then `npm run build:reference` gives error and tries to delete the user folder 
fixed it


davepagurek (approved):
Thanks for making a PR for this!

Issue #722, opened 2025-03-12
Vite is trying to compile the scripts/parsers/in contents


davepagurek:
### Most appropriate sections of the p5.js website?

Other (specify if possible)

### What is your operating system?

Windows

### Web browser and version

N/A

### Actual Behavior

When running `npm run dev`, sometimes we run into Vite errors because it's trying to compile the js code in `scripts/parsers/in` and you get an error message from the internals of p5.sound. @lukeplowden and I have also found that the time from starting the dev server to being able to load the site is noticeably faster (at least on windows) after deleting the contents of that folder.

### Expected Behavior

The `scripts/parsers/in` folder is mostly just used to produce the reference material. So far, we keep it around based on the assumption that cloning and setting up the main `p5.js` would be slower if we didn't, but we don't update the website reference all that often (just when a new version is released) compared to how often we make other changes to the site. So maybe it would be best if the reference parser script clears out the `scripts/parsers/in` folder after it's done? (Possibly `scripts/parsers/out` too?)

### Steps to reproduce

- [Build the reference](https://github.com/processing/p5.js-website/blob/main/docs/scripts.md#reference-build-script) from `npm run build:reference`
- Launch the dev server with `npm run dev`

### Would you like to work on the issue?

Someone else can take this on if desired!

kajal-jotwani:
I would like to work on this issue

webermayank:
Hey @davepagurek 
I have created the scripts for clearing out `src/parsers` content and it is working perfectly, `npm run build:reference` creating the reference files in `src\content\reference\en` and then deleteing the cloned repo in `src/parsers/in` and `out` too 
my question is that  i have created a `cleanUpDirectory` function in `src/scrips/utils.ts` , because the `cloneLibraryRepo` was also present in the utils.ts, is it the right place to setup that function ?

davepagurek:
I think that's fine @webermayank!

webermayank:
Okay i'll do some final tests then raise a PR

ksen0:
Should this be closed now that the PR is merged @davepagurek ? Or what remains to be done?

PR #731, opened 2025-03-17
Added scripts to clear parsers/in and parsers/out after build:reference

webermayank:
Fixes the issue #722 

The following changes had been done in this PR
1. Added a function `cleanUpDirectory` in `scripts/utils.ts` 
2.  Call that function in `reference.ts` which will be executed at the end of all the processes to clean the cloned `p5.js` and `p5.sound.js` repo in `scripts/parsers/in` and `out` too

after that dev server is starting faster without any errors

@davepagurek tell me if anything need changes, i would love to do so

### DIFF
--- a/src/scripts/parsers/reference.ts
+++ b/src/scripts/parsers/reference.ts
@@ -1,4 +1,4 @@
-import { cloneLibraryRepo, p5RepoUrl, readFile } from "../utils";
+import { cloneLibraryRepo,cleanUpDirectory, p5RepoUrl, readFile } from "../utils";
 import fs from "fs/promises";
 import { exec, execSync } from "child_process";
 import path from "path";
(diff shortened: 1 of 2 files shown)


davepagurek:
Nice, this looks good! Have you tried running `npm run build:reference` and then `npm run dev`? (Doing so will change a bunch of files in the reference folder that we don't want to commit just yet, so after testing you can run `npm run custom:cleanup` to clean those changes out.)

webermayank:
Yes @davepagurek  i tried `npm run build:reference` this has build the cloned repo along with some modifications in lots of files in `reference/en` directory files, and then `npm run dev` , the server started, and after that i did `npm run custom:cleanup`, the files that changes in `reference/en`  reverted back hence everything worked perfectly without any errors

davepagurek:
ok perfect, thanks for checking! This looks good to go then!

PR #743, opened 2025-03-22
Adding hwb.png for colorMode() docs

perminder-17:
I am updating the colorMode() docs, we know we have different colorModes, so to explain hwb colorMode() I am uploading an image for more clarity to the users. Here's the result how it looks. Here's the docs : https://github.com/processing/p5.js/pull/7654

Screenshot from 2025-03-22 05-44-50


perminder-17:
CC: @davepagurek

PR #745, opened 2025-03-23
Fixed scrollbar in example output frame

### LINKED ISSUE #716: Scrolls in example code output frame seems unnecessary

Reference

Windows

### Web browser and version

chrome://133.0.6943.143

### Actual Behavior

When we run the example codes  - 
https://p5js.org/reference/p5.Camera/centerX/ 
https://p5js.org/reference/p5/arc/ 
https://p5js.org/reference/p5/triangle/
etc
There comes an unwanted scroll in the small window, making it even more smaller and makes it hard to see what is actually happening 
- images 

Image

Image

### Expected Behavior

There shouldn't be any scrolls even after running the code if not necessary , the output frame should look same before and after pressing the run button like -

Image

Image

### Steps to reproduce

1. onto your chrome browser (this issue specifically happening on chrome)
2. go the urls mentioned above
3. locate to any example codes
4. click on run button - you can see the scrolls

yes

Its-sunny69:
Fixes #716 
Partially addresses processing/p5.js#7620


ksen0:
Hello @Its-sunny69 - so the `overflow: hidden` solution affects not only the case you showed, but also the example that @davepagurek is describing in [the comment on the related issue](https://github.com/processing/p5.js-website/issues/716#issuecomment-2749075705) that _should_ have scrollbars: https://p5js.org/reference/p5/input/ - that comment also suggests making the fix a bit more targeted:

> It seems like this issue also talks specifically about scroll bars for sketches that already have correctly-sized frames, but where scroll bars appear when zoomed in at certain levels. I think another valid solution could be to figure out how to remove scroll bars in just those cases, but still keeping scroll bars for the case where something is truly outside of the canvas and needs to scroll.

Could you update your solution to more specifically target your particular case, please?

Its-sunny69:
@ksen0 Thanks for providing more clarity on this issue, i will look for more specific solution .

package.json
```diff
@@ -35,6 +35,7 @@
     "js-yaml": "^4.1.0",
     "keyword-extractor": "^0.0.28",
     "lodash": "^4.17.21",
+    "p5.js-website": "file:",
```
ksen0, line 38: Could you leave out package and package-lock from the PR, like with `git ignore` please?

Its-sunny69: Yes

Its-sunny69:
Hey @ksen0 , if this PR solution is still valid and correct then please merge it.

doradocodes:
Closing this PR since the commit will be merged here: https://github.com/processing/p5.js-website/pull/1431

PR #762, opened 2025-03-29
resolves #702 added comments to all english mdx files.

### LINKED ISSUE #702: Should we add a message to reference mdx files saying they're auto-generated?
### Topic

It's a very common mistake for newcomers to this repo to try to edit a reference mdx file -- editing them does work, but they will get overwritten in the next release because the p5.js repo is the source of truth for (English) reference items.

Maybe we could add a comment to the top of each mdx file saying that it's auto generated? Ideally also with a link to where in the p5 repo it came from, similar to the link we put at the bottom of reference pages? This should only exist in the English ones, since the translations *do* actually live in this repo.

Swapnil-6161:
@davepagurek  Sir i have added the comment with the refence links of the pages they are sourced from. kindly look into my work.


ksen0:
Hi @Swapnil-6161 ! I'll close this PR because looks like it was already assigned to someone else before; if you want you can stay tuned on the original issue discussion in case actually it's open for work. Please note that Dave's comment (this one: https://github.com/processing/p5.js-website/issues/702#issuecomment-2661093567) specifies where the change should be. The /mdx files are auto-generated, so this fix should affect 1 file (the one in the comment). However, please help off from submitting PRs unless you have been assigned for the issue in the future. Thanks for understanding!

PR #795, opened 2025-04-14
adding XYscope.js to libs

ffd8:
[XYscope.js](https://teddavis.org/xyscopejs/) is the new p5.js port of [XYscope](https://teddavis.org/xyscope/) library for Processing. 

On Processing's library page it was added to `animation`/`hardware`/`sound` - since it touches all of those aspects. However the p5.js libs are only listed once. Therefore, I figured it's probably best under `drawing` although sound plays a really important role...


davepagurek:
Thanks @ffd8! Very cool thumbnail too.

@ksen0 We could probably switch to something like the [Processing libraries page](https://processing.org/reference/libraries/) where a library can be under multiple topics if we want and if it seems like there's interest.

Issue #810, opened 2025-04-17
Object properties not displaying correctly in reference pages


davepagurek:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Mac OS

### Web browser and version

Firefox

### Actual Behavior

On reference items that include objects with documented properties, e.g. https://beta.p5js.org/reference/p5.font/texttomodel/, the properties are showing up as if they're normal positional parameters:

![Image](https://github.com/user-attachments/assets/671d2916-bc00-4b74-a7bf-0aa6dea3c8a7)

### Expected Behavior

The properties should be excluded from the parameter list

### Steps to reproduce

Visit https://beta.p5js.org/reference/p5.font/texttomodel/

### Would you like to work on the issue?

Feel free to take this on!

kajal-jotwani:
Hello  @davepagurek, I would like to work on this issue

davepagurek:
Thanks @kajal-jotwani! I'll assign this to you.

kajal-jotwani:
@davepagurek  I had a question  do we need to completely remove the object properties from the parameters and syntax?

davepagurek:
Sorry for the delay, I ended up leaving a similar comment on the PR. but for continuity here: I think we want to still document the properties in the section where we list the parameters by name along with their descriptions, but not show them at all in the Syntax section.

NalinDalal:
hi, @ksen0 & @davepagurek , just to be on safe side, is this assigned to someone, or something like that?

ksen0:
Hi @NalinDalal thanks for the bump - looks like @kajal-jotwani's PR needs review: https://github.com/processing/p5.js-website/pull/818 I'll take a look later today and merge if possible, but please feel free to review this one!

NalinDalal:
hey @ksen0 any updates?

kitlord:
Hi @ksen0 @davepagurek I think the issue is fixed currently please close the issue if fixed

ksen0:
Thanks for following up on this @Iron-56 - yes looks like it's fixed. If it recurs, the thread can of course be reopened.

davepagurek:
Hey everyone, reopening this because I checked again and this criterion from above isn't being met still:

>  I think we want to still document the properties in the section where we list the parameters by name along with their descriptions, but not show them at all in the Syntax section.

[...]

PR #818, opened 2025-04-23
fix: prevent object properties from showing as separate parameters (#810)

kajal-jotwani:
Fixes: #810 

This PR fixes the issue where object properties were incorrectly showing up as separate parameters in the reference documentation.

image

### DIFF
--- a/src/scripts/builders/reference.ts
+++ b/src/scripts/builders/reference.ts
@@ -306,11 +306,42 @@ const convertToMDX = async (
 
 const getMethodFrontmatter = (doc: ReferenceClassItemMethod) => {
   const { params, return: returns, example, overloads, itemtype } = doc;
+  
+  // Process params to handle nested object properties
+  const processedParams = params?.map(param => {
+    // If this is a nested property (contains a dot), skip it
+    if (param.name.includes('.')) {
+      return null;
+    }
+    
+    // If this is an object parameter, collect its nested properties
+    if (param.type === 'Object' && params) {
+      const nestedProps = params
+        .filter(p => p.name.startsWith(`${param.name}.`))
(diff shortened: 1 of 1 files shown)


davepagurek:
Thanks @kajal-jotwani! Do you think there would be a way to still see a way to display the properties and their descriptions below the object itself in this example, but just not show it in the Syntax section as separate items? e.g. keeping this, which looks great:
image

...but below, showing the properties still? maybe something like this? (just a mockup, open to other designs!)
image

kajal-jotwani:
I think we can do that I'll try to do it and share the updates

kajal-jotwani:
image

Hello @davepagurek! 
I wanted to get your review on this whenever you get the chance.

ksen0:
Hi @davepagurek , it's been noted that the original issue https://github.com/processing/p5.js-website/issues/810 has been resolved - do you think this PR can still be useful?

Apologies for the long wait @kajal-jotwani , and thanks for your work on this.

davepagurek:
I think this is only partially resolved, as we don't show the properties of objects anywhere in the reference section of a docs page anymore, they're just hidden. So I think something like this is still worth doing. 

@kajal-jotwani this is definitely a step in the right direction, but as it is all the properties are getting squished into one line. I think to be mergeable, we'd want each property on its own line, indented in for readability.

kajal-jotwani:
@davepagurek Thanks for the feedback! I'll work on displaying each property on its own line and will share an update soon.

Nwakaego-Ego:
Hi @kajal-jotwani, just checking in on this. It has been a few months since the last update. Are you still working on the feedback from @davepagurek? 

CC @doradocodes @ksen0

PR #824, opened 2025-04-29
Updated reference regex to not break on beta versions

### LINKED ISSUE #808: Update reference regex to not break on beta versions

Reference

Mac OS

### Web browser and version

Firefox

### Actual Behavior

In the reference builder, we add a `v` to the start of the latest release if it seems to all be numeric: https://github.com/processing/p5.js-website/blob/df16dae34f923a743d6a589ab97603bf7a185723/src/scripts/parsers/reference.ts#L26-L28

This regex breaks for beta releases, e.g. `2.0.0-beta.5`. We're about to release 2.0 so we won't have beta releases for a while, but we may want this working for the future when we do beta 3.0 releases.

### Expected Behavior

The regex should match strings like `2.0.0-beta.5` too.

### Steps to reproduce

Manually edit this https://github.com/processing/p5.js-website/blob/df16dae34f923a743d6a589ab97603bf7a185723/src/globals/p5-version.ts#L1 to not start with a v

Feel free to take this on for me!

webermayank:
Updated reference regex to handle the version with  suffix like `1.2.3-beta`  too, eg-  `2.0.0-beta.8`.
fixes #808 
PR to main branch


webermayank:
@ksen0  The tests are now passing. Thanks for syncing main into PR branch! Let me know if there's anything else you'd like me to update — or if it's good to go for merging ?

PR #836, opened 2025-05-10
Support rest types in Syntax section of reference

davepagurek:
Some of our signatures now support rest types, e.g. createVector. We currently do not display or type check this correctly. I've made a PR here to support type checking https://github.com/processing/p5.js/pull/7803, and this adds the capability to display them in the reference.

It currently looks like this, using @GregStanton's idea borrowing from MDN the `x1, x2, ..., xn` syntax.
image

The `Number:` type is a little awkward here, but probably fine for now, and better than what we currently have.

#### JavaScript

If we want the syntax section to reflect how one would write this in your own function, it would be:
```js
createVector(...components)
```
...and also we'd document `components` as an *array*.

#### MDN

image

#### w3schools

image

#### Three.js

image

#### Lodash

image

#### Immutable.js

image

#### Oracle Java docs

image

### DIFF
--- a/src/components/MethodSignature/index.astro
+++ b/src/components/MethodSignature/index.astro
@@ -14,8 +14,13 @@ import { CopyCodeButton } from "../CopyCodeButton";
 
 const { params, name, overloads } = Astro.props;
 
-const formatParam = (param: ReferenceParam) =>
-  param.optional ? `[${param.name}]` : `${param.name}`;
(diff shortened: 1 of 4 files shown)


GregStanton:
This is fantastic @davepagurek! I love the list of documentation practices across the JS ecosystem and beyond :)

PR #847, opened 2025-05-20
submission for shapeSVG library

irenelfeng:
There is no SVG reading support in p5js according to: https://www.gorillasun.de/blog/working-with-svgs-in-p5js/, 
this library creates a loadSVG() function to load in an SVG and edit the style of each path in the XML file, and even individual vertices if need be.


ksen0:
Hi @irenelfeng thanks for working on this! Could you please update the image to use a different name? p5.svg is already in use by https://github.com/zenozeng/p5.js-svg . Once that's done, happy to merge!

ksen0:
Thanks @irenelfeng ! If this library is compatible with p5.js 2.0 (you can change to this version in the p5.js Editor to check), please feel free to make a PR like this into the 2.0 branch of this repository.

PR #863, opened 2025-06-09
p5.js Reference Translation Automation Task

Divyansh013:
Have implemented week 1 of my GSoC Project
The script detects if there is any change in english translation files and if any corresponing language files exists

### DIFF
--- a/.github/actions/translation-tracker/index.js
+++ b/.github/actions/translation-tracker/index.js
@@ -149,7 +149,7 @@ function displaySummary(translationStatus) {
   
   console.log('\n💡 Next steps for Week 2+:');
   
-
+}
 function exploreRepoStructure() {
   console.log('\n🔍 REPOSITORY STRUCTURE ANALYSIS');
   console.log('═══════════════════════════════════');
--- a/.github/actions/translation-tracker/index.js
+++ b/.github/actions/translation-tracker/index.js
@@ -1,0 +1,236 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+
+const SUPPORTED_LANGUAGES = ['es', 'hi', 'ko', 'zh-Hans'];
+const REFERENCE_PATH = 'src/content/reference';
+const ENGLISH_PATH = path.join(REFERENCE_PATH, 'en');
+
+
+function getChangedFiles() {
+  try {
+    console.log('🔍 Detecting changed files...');
+    
+    
+    const gitCommand = process.env.GITHUB_EVENT_NAME === 'pull_request' 
+      ? 'git diff --name-only HEAD~1 HEAD'
+      : 'git diff --name-only HEAD~1 HEAD';
+    
+    const changedFilesOutput = execSync(gitCommand, { encoding: 'utf8' });
+    const allChangedFiles = changedFilesOutput.trim().split('\n').filter(file => file.length > 0);
+    
+ 
+    const changedEnglishFiles = allChangedFiles.filter(file => 
+      file.startsWith(ENGLISH_PATH) && file.endsWith('.mdx')
+    );
+    
+    console.log(`📝 Total changed files: ${allChangedFiles.length}`);
+    console.log(`🇺🇸 Changed English reference files: ${changedEnglishFiles.length}`);
+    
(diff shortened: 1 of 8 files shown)


.github/actions/translation-tracker/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "p5js-translation-tracker",
+  "version": "1.0.0",
```
ksen0: Minor: what do you think about using "1.0" for the final working version of this; until then please keep as 0.1.0 (and increment with changes)?

.github/actions/translation-tracker/index.js
```diff
@@ -0,0 +1,228 @@
...
+const REFERENCE_PATH = 'src/content/reference';
+const ENGLISH_PATH = path.join(REFERENCE_PATH, 'en');
+
+
+function getChangedFiles() {
+  try {
+    
+    
+    
+    const gitCommand = process.env.GITHUB_EVENT_NAME === 'pull_request'
```
ksen0, line 18: To support local testing, I would suggest separating out this logic a little bit; ie, separate the logic of getting the list of affected files from the rest. The rest could be run on its own, but this is very integrated into GH and more difficult to test; or maybe you could include your testing setup in the description please?

.github/actions/translation-tracker/index.js
```diff
@@ -0,0 +1,228 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+
+const SUPPORTED_LANGUAGES = ['es', 'hi', 'ko', 'zh-Hans'];
+const REFERENCE_PATH = 'src/content/reference';
```
ksen0: I would recommend starting with tutorials or examples, since these exclusively live in this repository. For reference, [data.json](http://p5js.org/reference/data.json) is the structures overview, and may be needed; for example, if there is a code update (rather than update in the description or text), then maybe translation may not be needed?

.github/actions/translation-tracker/index.js
```diff
@@ -0,0 +1,228 @@
...
+    
+    return changedEnglishFiles;
+  } catch (error) {
+    console.error(' Error getting changed files:', error.message);
+    return [];
+  }
+}
+
+
+function getTranslationPath(englishFilePath, language) {
```
ksen0: This helper function is only used once - since it's already quite complex, I'd suggest not using a helper but using the one-line replace directly

.github/actions/translation-tracker/index.js
```diff
@@ -0,0 +1,228 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+
+const SUPPORTED_LANGUAGES = ['es', 'hi', 'ko', 'zh-Hans'];
+const REFERENCE_PATH = 'src/content/reference';
+const ENGLISH_PATH = path.join(REFERENCE_PATH, 'en');
```
ksen0: Very minor / same as below: if the constant is not reused, consider not creating it separately, so that code is more readable

.github/actions/translation-tracker/index.js
```diff
@@ -0,0 +1,228 @@
...
+  try {
+    const referencePath = REFERENCE_PATH;
+    console.log(`📁 Reference path: ${referencePath}`);
+    
+    if (fs.existsSync(referencePath)) {
+      const languages = fs.readdirSync(referencePath)
+        .filter(item => fs.statSync(path.join(referencePath, item)).isDirectory())
+        .filter(item => !item.startsWith('.'));
+      
+      console.log(`🌐 Available languages: ${languages.join(', ')}`);
```
ksen0, line 163: This overview is so nice! I really appreciate seeing it all at once for context.

src/content/reference/en/p5/circle.mdx
```diff
@@ -6,6 +6,8 @@ file: src/core/shape/2d_primitives.js
 description: >
   <p>Draws a circle.</p>
 
+  <p>This is a test modification for the translation tracker.</p>
```
ksen0: From these 3 files being here, I'd expect" Total changed files: 1" should be getting 3, but it's getting 1. What's the intended behavior?

PR #876, opened 2025-06-24
Update animating-with-media-objects.mdx - remove mention of preload()

ksen0:
Although the tutorial has been updated to use async/await, it still mentions preload() in one spot.Although the tutorial has been updated to use async/await, it still mentions preload() in one spot.


perminder-17 (approved):
This looks good to me!

Issue #873, opened 2025-06-24
[Image] Excessively long alternative text


coseeian:
### Title

[Image] Excessively long alternative text

### Description

An image (`<img>`) element on the page with very long (over 100 characters) alternative text. Non-text content must convey the same purpose or information as presented visually. When alt text is too long, users who rely on it may miss essential context or struggle to understand the content.

### Steps to Reproduce

1. go to tutorials/ and tutorials/setting-up-your-environment/
2. Locate to the following element
```
<img class="aspect-photo object-cover" height="293" width="440" src="/_astro/SetupA.MT9CUmVs_UdPSt.webp" loading="eager" decoding="async" alt="An interactive sketch in the p5.js Web Editor draws circles on the canvas as the mouse pointer moves." style="">
```
```
<img height="1152" width="2048" src="/_astro/p5_editor_sign_up_page.ti-nvYfB_Z17jU5E.webp" loading="lazy" infersize="true" decoding="async" aspectratio="none" alt="An arrow pointing to the highlighted User Name, Emaill, Password, and Confirm Password fields on the Sign Up page." style="">
```

### Actual Behavior

The image’s alt attribute contains more than 100 characters of description.

### Expected Behavior

Alternative text should be concise and focused, describing the essential content and function of the image in a brief phrase or sentence.

### Suggested Fix

1. Keep the alt text brief and focused—describe only the image’s essential purpose in a few words.
2. If readers need more context, place supplementary details next to the image or link to a separate description page.
3. Avoid hiding unique information solely in the alt text; ensure all critical content is visible elsewhere on the page.

### Reference

* [1.1.1 Non-text Content](https://www.w3.org/TR/WCAG22/#non-text-content)
* [F67: Failure of Success Criterion 1.1.1 and 1.2.1 due to providing long descriptions for non-text content that does not serve the same](https://www.w3.org/WAI/WCAG22/Techniques/failures/F67) purpose or does not present the same information
* [G73: Providing a long description in another location with a link to it that is immediately adjacent to the non-text content](https://www.w3.org/WAI/WCAG22/Techniques/general/G73)
* [G74: Providing a long description in text near the non-text content, with a reference to the location of the long description in the short description](https://www.w3.org/WAI/WCAG22/Techniques/general/G74)

### What is your operating system?

None

ghost:
Hey !
Can I work on this issue ?

ksen0:
[...]

PR #885, opened 2025-06-28
Changed excessively long alt text for 2 images 

ghost:
Addresses issue #873.


ksen0:
Thanks for your patience on this @sprinklehappiness! Could you please amend the PR to focus on just the 2 too-long texts? It looks like a third text was changed, but it was not too long.

For the other two, I think it would be better if the shortened version did not have less information. I'll add comments to each individually, but the following resources may be useful:

* [An alt Decision Tree](https://www.w3.org/WAI/tutorials/images/decision-tree/) more generally
* and specifically, [alt text for Complex Images](https://www.w3.org/WAI/tutorials/images/complex/)

From the above, if the image contains important information and the information is already present in surrounding text, you could use something like "as described above"  to make sure the removed information is still clearly available.

src/content/tutorials/en/setting-up-your-environment.mdx
```diff
@@ -4,7 +4,7 @@ description: A quick tutorial for setting up the p5.js Web Editor and VS Code to
 category: introduction
 categoryIndex: 0
 featuredImage: ../images/featured/SetupA.jpg
-featuredImageAlt: An interactive sketch in the p5.js Web Editor draws circles on the canvas as the mouse pointer moves.
+featuredImageAlt: An interactive sketch that draws circles as the mouse pointer moves.
```
ksen0: Seems like "An interactive sketch in the p5.js Editor draws circles on the canvas as the mouse pointer moves." would also be under 100 characters, and preferred, as it keeps more of the information

src/content/tutorials/en/setting-up-your-environment.mdx
```diff
@@ -73,7 +73,7 @@ If you are new to p5.js, start with the [p5.js Web Editor](https://editor.p5js.o
   - Create and confirm a password. 
   - Click the “Sign Up” button.
 
-![An arrow pointing to the highlighted User Name, Emaill, Password, and Confirm Password fields on the Sign Up page.](../images/introduction/p5_editor_sign_up_page.png)
+![An arrow to the highlighted fields on Sign-Up page.](../images/introduction/p5_editor_sign_up_page.png)
```
ksen0: E.g., "An arrow pointing to the highlighted fields on the Sign Up page, as described above."

PR #891, opened 2025-07-04
Fixes Screen reader and visual presentation are inconsistent. (issue#882)

### LINKED ISSUE #882: [Content] Screen reader and visual presentation are inconsistent.
### Title

[Content] Screen reader and visual presentation are inconsistent.

### Description

There is an inconsistency between the screen reader output and the visual presentation. After implementing random homepage images, only one image is displayed visually at a time, but all image links remain in the DOM and are focusable. This causes a mismatch between what sighted users see and what screen reader users encounter, leading to confusion and unnecessary navigation stops.

### Steps to Reproduce

1. Go to /
2.  Locate to the following element:
- body > div > header > div > a:nth-child(2)
- body > div > header > div > a:nth-child(10)

### Actual Behavior

Links corresponding to non-displayed images remain in the tab order and are announced by screen readers link role only, even though those images are hidden visually. A link that provides no information is focusable via the Tab key.

### Expected Behavior

Only the link for the image currently displayed should be focusable. Links for hidden images must be removed from the tab sequence and hidden from assistive technologies. The link cannot receive focus when the randomly displayed image does not correspond to it.

### Environments

_No response_

### Suggested Fix

Apply the hidden class or appropriate ARIA attributes to the <a> element itself, rather than to its child <div>. For example:
```
<!-- Before -->
<a href="...">
  <div class="hidden">...</div>
</a>

<!-- After -->
<a href="..." class="hidden" aria-hidden="true" tabindex="-1">
  <div>...</div>
</a>
```

This ensures that links for hidden images are not focusable or announced.

### Reference

WCAG:
* [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG22/#info-and-relationships)
* [2.4.11 Focus Not Obscured (Minimum)](https://www.w3.org/TR/WCAG22/#focus-not-obscured-minimum)

None

### Web browser and version

_No response_

coseeian:
Apply the hidden class or appropriate ARIA attributes to the element itself, rather than to its child

Resolves #882

### DIFF
--- a/src/components/PageHeader/HomePage.astro
+++ b/src/components/PageHeader/HomePage.astro
@@ -39,9 +39,9 @@ const { config } = Astro.props;
   {
     config.data.heroImages.map((im, i) => (
       im.linkTarget ?
-      <a href={im.linkTarget }>
+      <a href={im.linkTarget} class={`hero-image-container ${i > 0 ? "hidden" : ""}`}>
       <Image
-          containerClass={`relative hero-image-container ${i > 0 ? "hidden" : ""}`}
+          containerClass={"relative"}
           class={"hero-image"}
           aspectRatio="none"
           src={im.image}


ksen0:
Thanks @coseeian ! I'll go ahead and merge this. Because the `main` branch and `2.0` branch have diverged but are both active (one generates p5js.org, the other generates beta.p5js.org), I have a also cherry-picked it here: https://github.com/processing/p5.js-website/pull/904

PR #892, opened 2025-07-04
Fix: remove trailing slashes from broken Wikipedia links

### LINKED ISSUE #878: Links to wikipedia are broken

Reference

Windows

### Web browser and version

### Actual Behavior

### Expected Behavior

### Steps to reproduce

nope.

reshma045:
**Description**:
This PR fixes issue #878 by removing unnecessary trailing slashes from Wikipedia links across the p5.js website and documentation. Trailing slashes on some Wikipedia URLs were causing 404 errors.

**Issue**: Fixes #878

**Changes**:
Removed trailing slashes from all `wikipedia.org/wiki/` links found using a site-wide search.

**Example Fixed Link:**
Before: https://en.wikipedia.org/wiki/Screen_reader/ (404)
After: https://en.wikipedia.org/wiki/Screen_reader (Works (HTTP 200))

**Verification**:
- Verified changes on local server: `http://localhost:4321`
- Checked affected pages in both reference, contributor sections and other pages.
- Confirmed fixes across all five supported languages


reshma045:
Few unrelated tests (CodeEmbed.test.tsx, CodeFrame.test.tsx) failed due to a known syntax error (`prevRefreshReg` already declared). This failure is unrelated to the changes made in this PR, which only update markdown content and fix external Wikipedia links.

All reference and documentation pages were successfully built and verified locally.

blackboxlogic:
It looks like your change is to use old versions of two packages. I'm guessing the **add-a-slash** behavior was introduced in their newer versions?
Does this mean those packages can never be updated here? Is there something to remind people not to update them? Maybe a unit test, comment?
Was there a bug in those packages and has it been reported to the maintainers of them?

reshma045:
Hi @blackboxlogic thank you for the feedback!
Just to clarify, this PR does not modify or downgrade any dependencies or packages. The issue was purely in the markdown and JSON content. Several Wikipedia links had hardcoded trailing slashes (such as `/wiki/Some_Article/`) that led to `404`s. I performed a site-wide search and manually updated only the affected links in the content files (no logic or package behavior involved).
So: 1. No package versions were changed.
2. There is no lock-in to older versions.
3. There's no bug in a package that needs to be reported.
This change is isolated to the content layer and should remain stable regardless of future package updates.

blackboxlogic:
Look at your changed files, note that the first one included two package downgrades.

I wonder if you made manual updates to generated files, which I think would be an oops. I'm not familiar with this project, are you?

reshma045:
@blackboxlogic You're absolutely right, I initially didn't intend to change any dependencies, but I just noticed that `documentation` and `yuidocjs` were accidentally downgraded due to a local `npm install` resolving older versions (probably because of the lockfile or environment).

I’ve now restored the original versions:
`"documentation": "^14.0.3"`
`"yuidocjs": "^0.10.2"`

This PR only contains content fixes now, with no dependency changes. Thanks again for pointing this out!

blackboxlogic:
My question remains about if you made manual edits to generated content (removing all the slashes). I suspect the files you edited are the output from an automated process and changing them directly (while easy) is wrong.

But I'm not familiar with this project.

blackboxlogic:
My question remains about if you made manual edits to generated content (removing all the slashes). I suspect the files you edited are the output from an automated process and changing then directly (while easy) is wrong.

But I'm not familiar with this project.

blackboxlogic:
Also acknowledging that you're a stranger on the Internet who is trying to fix a problem and I appreciate that, and maybe you're familiar with this project and your change is correct and I shouldn't be so dismissive.

reshma045:
Thanks again, @blackboxlogic.
To clarify your question: no, I did not edit generated output files. All the changes were made directly in source markdown and JSON content files, such as `/src/data/reference/en.json`, `/src/content/contributor-docs/*.md` and their localized counterparts for other languages. The trailing slashes were hardcoded in these content files (not injected by a tool like `documentation.js` or `yuidocjs`). So the fix was safe to apply manually.
I'm not a core contributor yet, but I've spent time understanding this repo’s structure and verified the changes locally (`localhost:4321`) to ensure nothing broke.

limzykenneth:
This doesn't quite fix the problem as the contributor docs and reference pages are all generated content from their original in the p5.js repo. The original sources don't have trailing slashes on the links. The trailing slashes are added to enable relative links to work as expected in local development, which means they should not apply to absolute links such as wikipedia links. Without changing the adding slashes logic, the changes here will be replaced on next build. Therelevant changes needed are in [`src/pages/_utils-node.ts`](https://github.com/processing/p5.js-website/blob/90261edb2cb7126f0d00b60d007e5ee58d47890c/src/pages/_utils-node.ts#L74-L79) and [`src/scripts/builders/reference.ts`](https://github.com/processing/p5.js-website/blob/90261edb2cb7126f0d00b60d007e5ee58d47890c/src/scripts/builders/reference.ts#L207-L211), to add additional check that the URL does not start with `http` before adding the slashes.

There shouldn't be changes to `package-lock.json` as well. @reshma045 Let me know if you need any help.

reshma045:
Thanks for the clarification @limzykenneth.
 I’ll update the logic in `src/pages/_utils-node.ts` and `src/scripts/builders/reference.ts` to avoid adding slashes to absolute URLs, then rebuild and test and will share updates soon.

reshma045:
Hi @limzykenneth, @blackboxlogic , I wanted to provide an update on this issue.

What I have tried so far:
- Modifying `src/pages/_utils-node.ts` and `src/scripts/builders/reference.ts` to add checks preventing trailing slashes on absolute URLs, but the slashes still appear.
- Adding a post-processing step in the `convertToMDX` function to strip trailing slashes specifically from Wikipedia links before saving the MDX files, but this also didn’t prevent trailing slashes from showing up.

Despite these efforts, the trailing slash persists on absolute Wikipedia URLs in the built site.

Possible causes:
- The trailing slash might be introduced by another part of the build pipeline that I haven’t identified yet.
- There may be a global link rewriting or routing behavior that enforces trailing slashes.

I would appreciate any guidance or suggestions on where else I could look or how to properly fix this at the source. Thanks

davepagurek:
Just a quick check @reshma045, are you rebuilding with `npm run build:reference` before running the dev server?

reshma045:
Yes @davepagurek, I’ve been running `npm run build:reference` before starting the dev server with `npm run dev`, but the trailing slash still appears on Wikipedia links.

perminder-17:
Hi @reshma045, sorry I wasn’t involved in the earlier conversation and only just saw a few comments on this PR. I think I understand the issue you’re facing. I tried making the changes myself and got the correct results.

> Modifying src/pages/_utils-node.ts and src/scripts/builders/reference.ts to add checks preventing trailing slashes on absolute URLs, but the slashes still appear.

I believe the changes you’re making might not be correct. When you update the r`eference.ts` file, you just need to `run npm run build:reference` and this will automatically regenerate the files and remove the extra slashes.

A very nice clean solution might me for this problem would be.

1. Since many files have been changed, creating a large diff that’s harder to review, could you open a new PR just for this fix?

In that new PR, just change the file [src/scripts/builders/reference.ts](https://github.com/processing/p5.js-website/blob/90261edb2cb7126f0d00b60d007e5ee58d47890c/src/scripts/builders/reference.ts#L207-L211) (since utils files doen't generate the reference? @limzykenneth).

Once, you updated the file, you just need to run `npm run build:reference` and it will auto-generate the reference again with removed slashes ( you don't need to do that all manually). 

Review the changes by launching the site locally, and if everything looks good, push them in the new PR.

If the changes in the reference.ts still doesn't genereates the correct output, probably you can share me the changes you did, I can help on that. Thanks for your help on this :)

reshma045:
Hi @perminder-17, thank you for the detailed explanation. I’m still relatively new to the codebase and have learned a lot from your guidance. I’ll open a fresh PR, update only `src/scripts/builders/reference.ts`, run `npm run build:reference` to regenerate, and verify locally before pushing. If it still doesn’t work, I’ll share my changes here for review. Thanks again.

perminder-17:
> Hi @perminder-17, thank you for the detailed explanation. I’m still relatively new to the codebase and have learned a lot from your guidance. I’ll open a fresh PR, update only `src/scripts/builders/reference.ts`, run `npm run build:reference` to regenerate, and verify locally before pushing. If it still doesn’t work, I’ll share my changes here for review. Thanks again.

Sure, feel free to share the changes if it doesn't works.

reshma045:
Hi @perminder-17, I've created a new PR with all the suggested changes for the issue. Please check #934. Thank you.

PR #899, opened 2025-07-13
Update abs.mdx

iamniky:
Clarify abs() reference explanation


perminder-17 (approved):
Looks good to me! Thanks.

Issue #866, opened 2025-06-17
[Button] No feedback on action activation or completion for screen reader


coseeian:
### Title

[Button] No feedback on action activation or completion for screen reader

### Description

Assistive-technology(screen reader) users do not receive any perceivable feedback when the button activation and action completion.
When the button is activated by a keyboard or screen reader user, there is no focus change, nor is any feedback announced through a live region. As a result, assistive technology users cannot perceive whether the action was successful or not.

### page 1
1. go to /reference/p5/arc/
2.  Locate to the following element:
* Run sketch、Stop sketch、Copy code to clipboard、Reset code to initial value button
(This issue is since to be a global issue)

### page 2
1. go to /tutorials/setting-up-your-environment/
2.  Locate to the following element:
* Copy code to clipboard button(readonly textarea for code)
(This issue is since to be a global issue)

### Actual Behavior

1. Focus remains on the button with no change after the action completes.
2. No aria-live feedback or status message is announced after the action completes.

### Expected Behavior

moving focus to an appropriate element after the action completes.
2. An aria-live region should announce a concise status message after the action completes.

### Reference

WCAG:[Success Criterion 4.1.3 Status Messages](https://www.w3.org/TR/WCAG22/#status-messages)
WCAG-T:[ARIA22:Using role=status to present status messages](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22)
[ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions)

### What is your operating system?

None

ksen0:
Fixing PRs have been merged

PR #905, opened 2025-07-15
Add aria-live status messages for button actions to improve screen reader feedback (fixes issue #866)

coseeian:
This PR addresses issue #866.
After a button is pressed and the associated action is performed, a span container with aria-live is updated with a status message. This allows screen readers to announce feedback, ensuring that users receive confirmation when an action is completed.


ksen0:
This is great, thank you! My only question is about the timeout that is set in announce. As far as I can tell timeouts that are set are not cleared on unmount or when announce is called repeatedly. Does it make sense to add cleanup logic to handle potential race conditions?

cc @clairep94 @davepagurek

davepagurek:
I think that's right, maybe we need to keep track of the timer id, and clear the timer when we update the content + on unmount via a useEffect cleanup function?

coseeian:
I have already made the fix. Please review again.

PR #918, opened 2025-07-25
Prevent menu list from auto-expanding on Tab key navigation

### LINKED ISSUE #915: [Menu] Menu list automatically expands on Tab key navigation but is not visible
### Title

[Menu] Menu list automatically expands on Tab key navigation but is not visible

### Description

When using a mouse or keyboard to activate the "menu" button, the "menu list" panel expands and displays correctly.
However, when keyboard users navigate to the "menu" button (typically the third Tab stop) and then press the Tab key to move focus away, the "menu list" block automatically expands, but there is not enough space for it to be fully displayed on the screen. This prevents users from perceiving or interacting with the menu options, which poses a significant barrier for keyboard and assistive technology users.

### Steps to Reproduce

1. Navigate to `/` (the home page).
2. Switch to a mobile viewport (e.g., using device simulation in browser DevTools).
3. Use the Tab key to focus the "menu" button (usually the third Tab stop).
4. Press the Tab key again to move focus away from the "menu" button.
5. Observe that the "menu list" block automatically expands, but it is not fully visible due to insufficient screen space.

### Actual Behavior

When the "menu list" panel expands automatically after moving focus away from the "menu" button using keyboard navigation, there is not enough space to display the panel. As a result, some or all menu options may be hidden or inaccessible.

### Expected Behavior

The "menu list" panel should always be fully visible and accessible whenever it is expanded, whether triggered by mouse or keyboard click, or by automatic expansion. Its appearance and layout should match the behavior when the "menu" button is activated by mouse or keyboard clicking, with sufficient space to display all menu options.

### Environments

_No response_

### Suggested Fix

* **Option 1(recommend):Do not automatically expand the "menu list" panel when focus leaves the "menu" button; instead, only expand the panel when activated by mouse or keyboard click.
* **Option 2:Ensure that when the "menu list" panel is expanded automatically via keyboard navigation, its behavior matches that of expansion triggered by mouse or keyboard clicks, so the panel is always fully visible and accessible.

### Reference

_No response_

None

### Web browser and version

_No response_

AshmitSherigar:
### Problem
When keyboard users navigate using the `Tab` key, the menu list would auto-expand after the "Menu" button lost focus. On small viewports, this leads to the menu expanding off-screen, making it inaccessible for keyboard and assistive tech users.  
(See issue #915)

### Solution 
This PR ensures that the menu list **only expands when explicitly activated by click or keyboard press**, and **does not auto-expand** when focus leaves the "Menu" button. A keydown listener checks for `Tab` and closes the menu if focus moves outside.

### Changes Made
- Added a `useEffect` with a `keydown` event listener to detect when focus leaves the menu/button
- If the menu is open and the new focus is outside both, it closes the menu

### Accessibility Impact
- Improves accessibility for keyboard users by preventing unexpected or hidden UI behavior  
- Keeps focus management clean and predictable

---

Closes #915

### DIFF
--- a/src/components/Nav/MainNavLinks.tsx
+++ b/src/components/Nav/MainNavLinks.tsx
@@ -1,6 +1,7 @@
 import styles from "./styles.module.scss";
 import { Logo } from "../Logo";
 import { Icon } from "../Icon";
+import { useEffect, useRef } from "preact/hooks";
 
 type MainNavLinksProps = {
   links: {
@@ -26,6 +27,38 @@ export const MainNavLinks = ({
   isOpen,
   hasJumpTo,
 }: MainNavLinksProps) => {
+  //Fix : Menu list automatically expands on Tab key navigation but is not visible
+  const menuRef = useRef<HTMLDivElement>(null);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
+    const handleKeyDown = (e : KeyboardEvent) => {
+      if (e.key === "Tab"){
+        requestAnimationFrame(()=>{
+          const active = document.activeElement;
+          if (menuRef.current && buttonRef.current && !menuRef.current.contains(active as Node) && !buttonRef.current.contains(active as Node)) {
+
+            if (isOpen){
+              handleToggle();
+            }
+            
+          }
+        });
+      }
+    };
+
+    document.addEventListener("keydown",handleKeyDown);
+    
+
+  
+    return () => {
+      document.removeEventListener("keydown",handleKeyDown)
+    }
+  }, [isOpen , handleToggle])
+  
+
+
+
   if (!links || links?.length <= 0) return null;
 
   const renderLogo = () => (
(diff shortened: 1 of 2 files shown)


AshmitSherigar:
@coseeian could you review this PR

coseeian:
When the menu button's aria-expanded attribute is set to false, screen readers can still access the navigation items (Reference, Tutorials, Examples, Contribute, Community, About) by tab key or browser mode, even though the menu list is visually hidden. These navigation items should be hidden from the accessibility tree and tab key sequence when the menu list is not visible.

coseeian:
Additionally, in the desktop layout, pressing the Tab key from the top of the page, the third tab stop focuses an element that has no accessible name or label. It appears to be the menu button intended for the mobile layout. This button should also be completely hidden from both the accessibility tree and the tab sequence in desktop view.

AshmitSherigar:
@coseeian  could u check it now maybe

coseeian:
It works fine on mobile, but in the desktop layout, the third tab stop focuses an element with no accessible name or label—screen readers do not announce anything for this element. This element should be hidden from both the accessibility tree and the tab sequence in the desktop layout.

PR #936, opened 2025-08-14
Expand curated sketches to include 2025 set

MissTipo:
This PR updates the community curation logic and assets to integrate the 2025 curated sketches alongside the 2024 set.

**Changes:**

_src/api/OpenProcessing.ts_

- Removed hardcoded fetch limit to retrieve all curated sketches
- 
- Added support for fetching from both 2024 and 2025 curation IDs
- 
- Appended a curation property to each sketch for filtering and ordering

_src/pages/community.astro_

- Introduced priority ID list to select and order featured sketches
- 
- Prevented duplicate sketches from appearing in the display

_src/api/Images:_

- Added high-resolution images for all 2025 curated sketches

**Testing**

- Ran `npm run lint` `npm run check` — no linting errors.
- Ran `npm test` — all tests passed.
- Verified in local environment that:

  - All sketches load.
  - Priority sketches appear first with no duplicates.


src/pages/community.astro
```diff
@@ -3,8 +3,18 @@ import { getCurationSketches } from "../api/OpenProcessing";
...
+const allSketches = await getCurationSketches();
+const priorityIds = [
+  '2684408', '2693274', '2693345',
+  '2689119', '2484739', '2688829',
+  '2690571', '2690405', '2691712', '2690038'
+  ];
+const sketches = allSketches
+  .filter(sk => priorityIds.includes(String(sk.visualID)))
+  .filter((sk, index, self) =>
+    index === self.findIndex(s => s.visualID === sk.visualID)
```
davepagurek: What was the intention with this line? Just at a glance it looks like it may not be necessary on top of the previous filter.

MissTipo: Some sketches were duplicated when I still had limit specified here `const allSketches = await getCurationSketches(10);` so I added the second filter ensures no duplicate visualID entries appear in sketches, even if allSketches contains duplicates.

MissTipo: I left it  as a safeguard aginst duplication in the future even though the limit is not hardcoded right now.

davepagurek: Ahh, ok I see. Do you know at what point the duplicates are coming from? I wonder if it will cause other problems and make extra pages if `getCurationSketches()` is returning duplicates.

MissTipo: I logged the results from getCurationSketches() and confirmed it’s returning only unique visualIDs (54 total, no duplicates). The earlier duplication must have been caused by how the limit interacted with the priorityIds filter, so the extra duplicate filter isn’t needed now.
Screenshot from 2025-08-14 19-36-12

MissTipo: I’ve removed the extra duplicate filter.

davepagurek (approved):
Looks great, thanks!

PR #937, opened 2025-08-18
Fix broken webplatform.org links for AudioParam

md5:
I noticed that there were a number of broken links to documentation for `AudioParam` in the `p5.sound` docs, pointing to the defunct wiki on webplatform.org. This PR updates to point to [the `AudioParam` docs](https://webaudio.github.io/web-audio-api/#AudioParam) in the Web Audio spec instead.

There was an additional reference to the broken URL in `public/reference/data.json`, but it seems like that might be a generated file. I can update that reference as well if that's necessary or desirable.

### DIFF
--- a/src/content/reference/en/p5.Envelope/play.mdx
+++ b/src/content/reference/en/p5.Envelope/play.mdx
@@ -8,7 +8,7 @@ description: |-
   If the input is a p5.sound object (i.e. AudioIn, Oscillator,
   SoundFile), then Envelope will control its output volume.
   Envelopes can also be used to control any <a href="
-  http://docs.webplatform.org/wiki/apis/webaudio/AudioParam/">
+  https://webaudio.github.io/web-audio-api/#AudioParam">
   Web Audio Audio Param.</a></p>
 line: 5078
 isConstructor: false
--- a/src/content/reference/en/p5.Envelope/triggerAttack.mdx
+++ b/src/content/reference/en/p5.Envelope/triggerAttack.mdx
@@ -8,7 +8,7 @@ description: |
   Similar to holding down a key on a piano, but it will
   hold the sustain level until you let go. Input can be
   any p5.sound object, or a <a href="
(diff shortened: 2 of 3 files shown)


src/content/reference/en/p5.Envelope/play.mdx
```diff
@@ -8,7 +8,7 @@ description: |-
   If the input is a p5.sound object (i.e. AudioIn, Oscillator,
   SoundFile), then Envelope will control its output volume.
   Envelopes can also be used to control any <a href="
-  http://docs.webplatform.org/wiki/apis/webaudio/AudioParam/">
+  https://webaudio.github.io/web-audio-api/#AudioParam">
```
perminder-17, line 11: For eg, This part is auto generated from here  : https://github.com/processing/p5.js-sound/blob/6feca5f809b69a0c6b7fbee4d376e76d4c5e4abc/src/envelope.js#L403

perminder-17 (changes requested):
Hi @md5, thanks a lot for catching all these broken links and for fixing them. The only issue is that the p5.Sound docs are auto-generated from the [p5.sound repo](https://github.com/processing/p5.js-sound). So the best approach would be to make the changes directly in that repo instead. What you think?

src/content/reference/en/p5.Envelope/play.mdx
md5: Sure, I don't mind opening a PR against another repo. I looked around a bit to find the source and found this repo 😄 

One thing that might be nice is if the generated documents had a link back to the source to know how to submit fixes. Not sure if that would cause more trouble than it's worth.

md5: Opened https://github.com/processing/p5.js-sound/pull/808

md5:
Closed in favor of https://github.com/processing/p5.js-sound/pull/808

PR #943, opened 2025-08-22
Updated donate widget with new campaign information

ksen0:
As there is a new PF fundraising campaign, the embedded widget is updated. Will look like this:

image

Once this is merged to 2.0, can also cherry-pick to main.


xinemata:
Thanks Kit! It'd be nice to tie the [fundraiser message](https://donorbox.org/back-to-school-805292) to the form, but it doesn't look like Donorbox lets me add hyperlinks on the embed form. Would it make sense to replace the message on the donate page with the Software as an Act of Love fundraiser copy during the fundraiser period (Aug 25~Sep 30, 2025)?

@raclim has also started to work on translating the copy on [#3578](https://github.com/processing/p5.js-web-editor/issues/3578) which may be helpful.

ksen0:
I've updated the message, and @raclim I'll look into updating the translations accordingly for the other languages.

@xinemata from October, 2025, though, do you think it makes sense to use the previous text again, and/or keep all/some of the new text?

ksen0:
PS @davepagurek if you have a moment, any ideas on dealing with the failed checks on scrolling and frameborder? I can remove those to deploy, but it does look better this way, so if there's a way to ignore the check (safely), that would be nice.

_EDIT: Resolved_

src/components/DonorBoxWidget/index.astro
```diff
@@ -1,14 +1,19 @@
-<script src="https://donorbox.org/widget.js"></script>
+<!-- @ts-expect-error  -->
+<script src="https://donorbox.org/widget.js" paypalExpress="true"></script>
```
perminder-17: Actually the test is failing because of the `paypalExpress = "true"`.

here's the error:

> See docs for more details: https://docs.astro.build/en/guides/client-side-scripts/#script-processing.
Add the `is:inline` directive explicitly to silence this hint.
2 <script src="https://donorbox.org/widget.js" paypalExpress="true"></script>
         ~~~
src/components/DonorBoxWidget/index.astro:2:46 - error ts(2322): Type '{ src: string; paypalExpress: string; }' is not assignable to type 'ScriptHTMLAttributes & AstroScriptAttributes & AstroDefineVarsAttribute'.

Actually the browsers ignore unknown `<script>` attributes, So I think PayPal stuffs must be turned on/off in the UI and should be removed from here? @davepagurek ?

Also, In the documentation https://docs.astro.build/en/guides/client-side-scripts/#unprocessed-scripts it says, I think typecheck error was because `paypalExpress` isn’t a valid attribute?

ksen0: Thanks @perminder-17 ! You're right

PR #952, opened 2025-09-02
Auto Close issues

NalinDalal:
Resolves issue

Changes:
introduced yml file to auto close issue

[checkout for reference](https://github.com/processing/p5.js/pull/8025#issuecomment-3243225399)


ksen0:
Hi @NalinDalal , same feedback here as on the other PR, and also looks like the base is not correct. Please close this PR and open on that is based on `main` in the first place, and contains the yml with all the feedback included. As I mentioned before, it might be helpful if you do these **one at a time** (so, after the other one is corrected and merged) do avoid duplicate effort. Thank you!

NalinDalal:
ohh, okay makes sense

PR #895, opened 2025-07-06
[GSOC 2025]: Github action for translation tracker

Divyansh013:
- test-local file detects local changes
- index.js does manual triggering for issue creation
- action works on all "push" events
- outdated banner addition on missing translation pages

PR #953, opened 2025-09-02
[GSOC 2025] : Action only Pull Request 

Divyansh013:
In continuation of #895


Divyansh013:
@ksen0 have removed ui part from this PR, let me know if this is the correct way to do so

Issue #955, opened 2025-09-04
Build fails due to OpenProcessing rate limits (429)


perminder-17:
### Most appropriate sections of the p5.js website?

Home

### What is your operating system?

Linux

### Web browser and version

Firefox

### Actual Behavior

Our site build sometimes fails when pages fetch data from OpenProcessing. The API returns 429 Too Many Requests, which makes our code expect arrays/strings but get nothing, causing crashes.

- CI builds fail randomly.
- Unable to build the website's reference from the forked repo. 

Here's the CI failure : https://github.com/processing/p5.js-website/pull/954 when I tried uploading images. You can see the errors in the PR https://github.com/processing/p5.js-website/actions/runs/17451097533/job/49557683435?pr=954

```
getSketch 2215521 429 Too Many Requests
Cannot read properties of undefined (reading 'toLowerCase')
```

### Expected Behavior

There should be no crashes when OP returns 429 or bad JSON.

### Would you like to work on the issue?

No, I am not very confident with typeScript and CI testings, but a possible solution would be (during tests and local builds, don’t call OpenProcessing at all)

ksen0:
Thanks for logging this @perminder-17 (also cc @davepagurek since this maybe a repeat from this potentially related prior issue: https://github.com/processing/p5.js-website/issues/955)
I've marked is as "Critical" because it blocks many outstanding PRs.

I will investigate the following: (1) not doing excessive OP API calls in CI (just 1 confirm it works) and (2) ensure there's a reasonable fallback on the site itself.

davepagurek:
Based on the code in https://github.com/processing/p5.js-website/pull/838/files, it looks like it should only hit that error if we're trying to load a sketch that's not in one of our default curations which we load in bulk once. It looks like this could happen if:
- We are calling it with different params so it's not cached. Looks like we always call with the same params though.
- Something about `memoize` isn't working, e.g. we're building pages in parallel where it used to be serial and the separate processes aren't using the same cache. We could try having a manual build script that pulls the data and writes the result to a json file that we check into git, so that we aren't constantly querying OP on every build
- We're requesting a sketch that isn't in a curation (not sure where this would come from in our source code though)
- Something changed on OP's side. Do we see the failures starting after a change on our end?

I don't know what the specific issue is, but the second bullet would probably address it regardless if we make querying OP a manual step. If we aren't sure what the cause is, that could be a way forward? e.g.:
[...]

PR #971, opened 2025-09-12
Improve OpenProcessing API error handling and add caching tests

junseok44:
addresses #955 and is a follow-up to #959.

This PR improves error handling in the OpenProcessing API functions and adds tests, focusing on caching behavior, to prevent silent build issues.

## Main changes

- Added a validation filter in getStaticPaths() to exclude invalid sketch IDs.
- Changed console.error to throw Error in getCurationSketches, getSketch, and getSketchSize to ensure failures are not ignored.

Previously, when OpenProcessing API calls failed, we would only log the errors and continue the build process with potentially broken data. This could lead to successful builds with broken sketch pages or cause confusing build failures much later in the pipeline. with the changes the build fails immediately and explicitly when an API call fails to make debugging much clearer and prevent broken pages from being deployed.

- Added extensive tests for caching behavior and various error scenarios.

A test case for caching failures due to incorrect input types (e.g., passing a string to getSketch like #955) was intentionally omitted, as this type of error is better caught by static type checking.

## Notes

With these changes, if an API call for an individual sketch fails, the entire build will now fail. We could adjust this in the future to skip only the problematic pages, but failing fast seems like a safer approach for now.

We should also consider making the CI build conditional on tests passing, especially since we are now testing caching behavior that is critical for build performance.


ksen0:
Thank you @junseok44 for this excellent PR, the test suite is fantastic!
I'll merge this in but please feel free to respond below:

> We should also consider making the CI build conditional on tests passing, especially since we are now testing caching behavior that is critical for build performance.

Unless I'm misunderstanding, if these tests fail, this will already stop the workflow and prevent deployment / will be clearly visible. Did you mean something additional as well? I agree this is critical functionality.

junseok44:
Thanks for reviewing! As you said, the deployment won't proceed if a test fails.
My concern was that the build job, running in parallel, could make excessive API calls (overfetching due to unexpected caching behavior) before the cache-test fails and stops the workflow. Although this is unlikely since the test usually runs faster than the build, it's a hassle to wait at least 10 minutes even if we get blocked by overfetching. so i thought making build job run after the test job would be better option in this case.

ksen0:
Thanks for explaining @junseok44 , that seems totally reasonable, please feel free to open a PR adding the `needs: test` line to https://github.com/processing/p5.js-website/blob/main/.github/workflows/test.yml if you have time (referencing the original issue and this PR for context - since it's a followup)

junseok44:
sure, i've just added dependency on build in the PR above. please have a check!

PR #981, opened 2025-10-02
Add auto-generated comments to reference MDX files

### LINKED ISSUE #702: Should we add a message to reference mdx files saying they're auto-generated?

adi-ray:
Fixes #702 

This implements message in English reference MDX files to prevent accidental edits.


src/layouts/ReferenceItemLayout.astro
```diff
@@ -283,13 +283,12 @@ const descriptionParts = description.split(
...
            <div class="text-body [&_a]:text-type-magenta-dark [&_a]:!decoration-type-magenta-dark my-lg">
-            Notice any errors or typos? <a href ="https://github.com/processing/p5.js/issues">Please let us know</a>. Please feel free to edit 
+            This page is generated from the comments in
             <a
                 href={`https://github.com/processing/p5.js/blob/v${p5Version}/${entry.data.file}#L${entry.data.line}`}     
                   >
                   {entry.data.file}
-                  </a>
-                and open a pull request!
+                  </a>. Feel free to edit it and submit a pull request!
```
ksen0: I think it's alright to keep the original phrasing, "Please feel free to edit it and submit a pull request!"

ksen0:
Thanks for finishing this one @adi-ray ! Could you add a commit please with all the changes from `npm run build:reference` as well? The reference generation is typically triggered using the `p5.js` release workflow (https://github.com/processing/p5.js/blob/main/.github/workflows/release-workflow.yml) rather than on PR merge / deployment of the website repo. So this way the comment can also be included even before the next release.

Also, if you have time, please feel free to make the matching PR in `2.0` branch, cherrypicking these 2 commits there first and then running the autogeneration.

davepagurek:
> Also, if you have time, please feel free to make the matching PR in `2.0` branch, cherrypicking these 2 commits there first and then running the autogeneration.

One small clarification, on the 2.0 branch we'll want to run `npm run build:reference` + commit the results separately from on the `main` branch because the reference items are different. So for 2.0 you'll want to cherrypick the updates to the `src` directory you've made, but not the commit where you update the reference items, and then do a new commit after regenerating the docs 🙂 sorry it's kinda complicated, let us know if you need any help with any of that!

adi-ray:
> Could you add a commit please with all the changes from `npm run build:reference` as well?

image
image

Getting this error

ksen0:
@adi-ray Hmm, could you try the below please?

1. Remove the cloned library `rm -rf src/scripts/parsers/in/p5.js`
2. Run `npm run build:reference` again

I only get the above error when I run the build:reference script, then quit the process unexpectedly, and run it again. Otherwise, it has no errors. The filepath of the cloned library I could figure out from here: https://github.com/processing/p5.js-website/blob/e1d02aafd4107c2e808ebe94ea57b4d42ceb5d85/src/scripts/utils.ts#L30

If you're interested, you are welcome to investigate this issue and add some improved cleanup in `utils.ts` in a separate issue/PR. But if it doesn't work please let me know what happens when you attempt the cleanup.

adi-ray:
Thanks @ksen0! That was exactly the issue. After the cleanup, it worked perfectly!

adi-ray:
Running `npm run build:reference` is also modifying the file path separators in `data.json`
image

ksen0:
@adi-ray That one I can't reproduce at the moment on this branch. If you are on a Windows machine, it might be a path normalization issue when `data.json` is built, which wouldn't affect CI, but would still be worth making a little more easily runnable for future contributors. If you're not on Windows, then definitely it's worth to investigate further and I'm not sure what the underlying cause is, but I can take a look again in a few days! Either way, you're welcome to suggest a fix for the build:reference script

ksen0:
Sorry for the notification noise, it was a mistake! Merging this in shortly

PR #986, opened 2025-10-06
fix: prevent menu links from being focusable when menu is closed

### LINKED ISSUE #915: [Menu] Menu list automatically expands on Tab key navigation but is not visible

hxrshxz:
## Description

When keyboard users navigated to the menu button and pressed Tab to move focus away, the menu list would automatically expand but was not fully visible on mobile viewports. This occurred because the menu links were still in the tab order even when the menu was closed, causing focus to move inside the hidden menu panel.

Fixes #915 

## Solution

Added dynamic `tabIndex`  to navigation links:
- When menu is **open**: `tabIndex={0}` (links are focusable)
- When menu is **closed**: `tabIndex={-1}` (links are removed from tab order)

Menu list no longer automatically expands when keyboard users tab away from the menu button.

This makke sures that tabbing away from the menu button moves focus to the next element outside the menu, not to hidden links inside it.


hxrshxz:
@coseeian Please take a look

src/components/Nav/MainNavLinks.tsx
```diff
@@ -75,21 +75,31 @@ export const MainNavLinks = ({
...
+              {link.label}
+            </a>
           </li>
         ))}
       </ul>
       <ul class="flex flex-col gap-[15px]">
         <li>
-          <a className={styles.buttonlink} href="https://editor.p5js.org">
+          <a
+            class={styles.buttonlink}
```
coseeian: nit: the `className` to `class` changes aren't necessary for this PR's core functionality (the tabIndex changes). To keep the diff focused, we could leave these as `className` for now.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -75,21 +75,31 @@ export const MainNavLinks = ({
...
             <div class="mr-xxs">
               <Icon kind="code-brackets" />
             </div>
             {editorButtonLabel}
           </a>
         </li>
         <li>
-          <a className={styles.buttonlink} href="/donate/">
+          <a
+            class={styles.buttonlink}
```
coseeian: nit: the `className` to `class` changes aren't necessary for this PR's core functionality (the tabIndex changes). To keep the diff focused, we could leave these as `className` for now.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -75,21 +75,31 @@ export const MainNavLinks = ({
       <ul>
         {links.map((link) => (
           <li key={link.label}>
-            <a href={link.url}>{link.label}</a>
+            <a href={link.url} tabIndex={isOpen ? 0 : -1}>
```
coseeian: minor: Since `<a>` elements are naturally focusable, we only need tabIndex when hiding them. Consider using `tabIndex={isOpen ? undefined : -1}` instead of the ternary with 0, which makes the intent clearer and avoids redundant attributes.

This would apply to all three link groups (nav links on L78, editor button on L89, and donate button on L101).

src/components/Nav/MainNavLinks.tsx
coseeian: nit: the `className` to `class` changes aren't necessary for this PR's core functionality. To keep the diff focused, we could leave these as `className` for now.

src/components/Nav/MainNavLinks.tsx
coseeian: nit: the `className` to `class` changes aren't necessary for this PR's core functionality. To keep the diff focused, we could leave these as `className` for now.

src/components/Nav/MainNavLinks.tsx
coseeian: minor: since `<a>` elements are naturally focusable, we only need `tabIndex` when hiding them. Consider using `tabIndex={isOpen ? undefined : -1}` instead of the ternary with 0, which avoids redundant attributes in DOM.

This would apply to all three link groups (nav links on L78, editor button on L89, and donate button on L101).

coseeian (commented):
Thanks for your work on this.

For a more robust accessibility solution, I'd suggest using the `hidden` attribute on the container rather than managing `tabIndex` on each link.

```jsx
<ul id="main-links-list" hidden={!isOpen}>
  {links.map((link) => (
    <li key={link.label}>
      <a href={link.url}>{link.label}</a>
    </li>
 ))}
</ul>
```
(The snippet above is a conceptual example; you'll need to adapt it to fit your specific implementation.)

The key benefit is that `hidden` removes the elements from both the tab order and the accessibility tree. This is crucial for preventing a confusing experience where a screen reader might accidentally announce links that are supposed to be visually collapsed or not interactive.

Since `.mainlinks` doesn't use CSS transitions, using `hidden` won't negatively impact animations.

Additionally, to complete the semantics, we could add `aria-controls` to the toggle button. This explicitly links the controller to the content it manages, providing better context for assistive technology.

References:
* [HTML hidden attribute (MDN)](https://www.google.com/search?q=%5Bhttps://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden%5D(https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden))
* [aria-controls attribute (MDN)](https://www.google.com/search?q=%5Bhttps://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls%5D(https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls))

What are your thoughts on this approach?

coseeian:
Hi @hxrshxz ,

Just checking in to see how you're doing with this PR, as it's been a little while. I've left a few comments on the PR previously, and I'd welcome any discussion or thoughts you have on those!

Also, please let me know if you're still planning to continue - a quick confirmation within the next 7 days would be great. That's not a deadline to finish the task, just a ping to confirm you are still working on it. If not, no problem at all - the issue can be opened back up for other volunteers.

Thank you!

hxrshxz:
@coseeian Really sorry for the delay , I will update the PR by tomorrow I was a bit busy with some hackathons and university work :sweat_smile: 
thanks for your time

coseeian:
@hxrshxz no problem, take your time.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -72,24 +73,34 @@ export const MainNavLinks = ({
       }`}
     >
       {renderLogo()}
-      <ul>
+      <ul id="main-links-list" hidden={!isOpen}>
         {links.map((link) => (
           <li key={link.label}>
-            <a href={link.url}>{link.label}</a>
+            <a href={link.url} tabIndex={isOpen ? undefined : -1}>
```
coseeian: We can simplify this. Because `hidden` normally handles focus prevention, we can omit the `tabIndex` manipulation on each link.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -72,24 +73,34 @@ export const MainNavLinks = ({
...
+              {link.label}
+            </a>
           </li>
         ))}
       </ul>
-      <ul class="flex flex-col gap-[15px]">
+      <ul id="main-cta-list" class="flex flex-col gap-[15px]" hidden={!isOpen}>
```
coseeian: Per `hidden` attribute's HTML spec:

> Because this attribute is typically implemented using CSS, it's also possible to override it using CSS. For instance, a rule that applies 'display: block' to all elements will cancel the effects of the Hidden state.

When both hidden and display:flex appears on the same element (plus if we remove the manual `tabIndex={isOpen ? undefined : -1}` manipulation on each link for code simplicity), the element is still focusable.

To maintain proper compliance with the hidden attribute's intended behavior, I would suggest applying the `flex` class only when `isOpen=true`.

ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#the_hidden_state

coseeian (commented):
Thanks for the thorough work here! I've added some comments concerning a few small issues that need to be resolved before merging. Please address the feedback points, and happy to discuss if you have thoughts on those.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -72,24 +73,34 @@ export const MainNavLinks = ({
...
+        id="main-cta-list"
+        class={`${isOpen ? "flex" : ""} flex-col gap-[15px]`}
+        hidden={!isOpen}
+      >
         <li>
-          <a className={styles.buttonlink} href="https://editor.p5js.org">
+          <a
+            className={styles.buttonlink}
+            href="https://editor.p5js.org"
+          >
```
coseeian: nit: These lines appear to only have formatting changes without any functional modifications. Please revert to the original code to keep the diff clean.

src/components/Nav/MainNavLinks.tsx
```diff
@@ -72,24 +73,34 @@ export const MainNavLinks = ({
...
             </div>
             {editorButtonLabel}
           </a>
         </li>
         <li>
-          <a className={styles.buttonlink} href="/donate/">
+          <a
+            className={styles.buttonlink}
+            href="/donate/"
+          >
```
coseeian: nit: These lines appear to only have formatting changes without any functional modifications. Please revert to the original code to keep the diff clean.

coseeian (approved):
Others LGTM.

coseeian:
Hi @ksen0,

I've tested these changes locally and can confirm they resolve the issue reported in https://github.com/processing/p5.js-website/issues/915.

When you have a moment, would you mind reviewing the implementation <del>(aside from the minor issues mentioned above, which I think @hxrshxz will handle)</del>?

Thanks for your time!

coseeian (approved):

src/components/Nav/MainNavLinks.tsx
hxrshxz: Done

src/components/Nav/MainNavLinks.tsx
hxrshxz: Done

coseeian (approved):
LGTM, thanks. Let's just wait for p5.js Project Lead to weigh in.

PR #997, opened 2025-10-11
Fix: Remove trailing slashes from external URLs to prevent 404 errors

### LINKED ISSUE #878: Links to wikipedia are broken

hxrshxz:
Fixes #878 
## Description 
Wikipedia and other external links were breaking with 404 errors due to Astro's `trailingSlash: "always"` config which was appending slashes to all URLs.

## Changes
- Modified middleware to remove traling slashes from external URLs (except root domains like `https://example.com/`)
- Make sure that the middleware runs for all locales, including the default English local

https://github.com/user-attachments/assets/bd37cdca-5fc3-4ff1-bc06-85b53774ec06


hxrshxz:
@ksen0 @davepagurek Please take a lookk

hxrshxz:
@davepagurek please have a look when u get time , thank you

limzykenneth:
This is not quite the right fix, rather than removing the slashes after they have been added. The slashes should not be added in the first place. This can be achieved by including a check for links starting with the `http` protocol when [building the reference](https://github.com/processing/p5.js-website/blob/main/src/scripts/builders/reference.ts#L208-L212).

PR #1008, opened 2025-10-19
Simplify translation sync workflow triggers

ksen0:
Reduced paths and triggers so the workflow focuses on examples and tutorials.


ksen0:
Hi @Divyansh013 , because this was [recently failing](https://github.com/processing/p5.js-website/actions/runs/18620859818/job/53091824127), I'm updating the workflow to focus only on the examples and tutorials. Please feel free to create a PR to update the scope again - not sure if the error is related

PR #841, opened 2025-05-13
Detailed example attribution

ksen0:
**THIS PR WAS ACCIDENTALLY CLOSED BEFORE COMPLETION. REOPENED HERE: https://github.com/processing/p5.js-website/pull/1009/**

Addresses #415 - work in progress. Remains to be done:

- [x] Support for cases where pre-2023 attribution is also collective [e.g., Shape primitives](https://archive.p5js.org/examples/form-shape-primitives.html)
- [x] Support for cases where there is no link for author
- [ ] Once done on main, will need to be cherrypicked to 2.0 as well
- [ ] Ensure all people mentioned in any part of attribution is on all-contributors in p5.js README in both main and dev-2.0 branches and include "example" emoji. **Complete list:** `Prof** WM Harris`, `GH/prashantgupta24`

# Curent Behavior

The license is at the bottom of the page, and attribution (inspired by and maintained by) is split.

Screenshot 2025-05-13 at 12 21 45

(Then there is the sketch and the code - viewer has to scroll all the way down)

Screenshot 2025-05-13 at 12 21 51

# PR Behavior

Considering showing the attribution more clearly. "p5.js Contributors" also points to all-contributors list in p5.js repo README, rather than the People page. The screenshots includes rephrasing to increase clarity for each example but make the history more maintainable overall using a clearer schema; this is also open to feedback.

image

image

# Future use

Although attribution of p5.js contributors work is collective, this schema allows detailed attribution of referenced work from outside the community. For example, including the following in the `.mdx` of an example:

```
remix:
  - attributionLabel: Processing example by Ira Greenberg
    attributionURL: https://processing.org/examples/reflection1.html
    remixLabel: Based on
  - attributionLabel: David Blitz
    remixLabel: Ported by
  - attributionURL: https://github.com/dkessner
    attributionLabel: Darren Kessner
    remixURL: https://github.com/processing/p5.js-example
    remixLabel: Revised in 2023

collectivelyAttributedSince: 2024
```

Will create the following attribution:

image

Thus, new examples can be added that correctly attribute referenced material. In the above schema, the `collectivelyAttributedSince` field is applicable mainly to examples ported from the Processing or archived p5.js site. For future additions, the `remix` filed can be used to provide a complete history. The URLs for attribution and type of remix can be edited, with optional URLs. If there is no remixLabel, "Remixed from" will be used, but other phrasing can be provided for clarity.

Issue #415, opened 2024-06-19
License issue?


calebfoss:
I recently communicated concerns regarding attribution for the examples on this site. The public side of that conversation took place on #386. Most recently, I opened #412 as a temporary solution. 

I realized, however, that there could potentially be an issue with the [license](https://github.com/processing/p5.js-website-legacy/blob/main/LICENSE-MIT) on the legacy site repo, and I do not know if my PR would solve it. I have a difficult time understanding the legal language of licenses, and I wanted to raise the question for anyone who better understands licensing. My understanding is that with Github's [inbound=outbound](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license) policy, contributors' work on the legacy site would be licensed under the same [license](https://github.com/processing/p5.js-website-legacy/blob/main/LICENSE-MIT) as the repo and therefore also copyrighted. Because their code is now in this repo but with their authorship removed and replaced with a different author and because this repo does not currently include the required copyright notice or license, I am wondering if that is a violation of the legacy repo license.

calebfoss:
I am seeing a similar issue with the Reference. I do not see a way to trace that writing back to its authors.

For the sake of open source contributors in general, I hope that the MIT License does require attribution for the authors (via copyright notice I expect). Even if not, however, I would strongly advocate for attribution regardless. STF project workers like me were compensated and will be acknowledged in a planned social media post on site launch, but my understanding is that much of the work we were building off of was uncompensated with no plan to acknowledge those authors for the new site.

Qianqianye:
We initially planned using the same MIT license as the [legacy website repo](https://github.com/processing/p5.js-website-legacy?tab=MIT-1-ov-file#readme). I have been doing research on this and will get back on this issue after I gather more suggestions from previous contributors.

Qianqianye:
Hi @outofambit @stalgiag, just to double check, are there anything we need to aware about licensing from Bocoup's perspective (the main developer team of the new website)? Thanks.

calebfoss:
Thank you for the update, @Qianqianye! In the meantime, I opened #425 to add attribution links to each example's description. This replaces the placeholders I opened in #412. I briefly described my process in the PR. 

You could follow a similar process for the reference. It would be a bit more complicated, since it would involve matching up the JSdoc from the p5 library. It took me longer than I expected to get the example attribution together, so I'm going to pause there for now. I'm happy to answer questions about my process!

stalgiag:
[...]

PR #1009, opened 2025-10-22
Detailed example attribution

ksen0:
(Reopening https://github.com/processing/p5.js-website/pull/841, which was incorrectly / accidentally closed)

Addresses #415 - work in progress. Remains to be done:

- [x] Support for cases where pre-2023 attribution is also collective [e.g., Shape primitives](https://archive.p5js.org/examples/form-shape-primitives.html)
- [x] Support for cases where there is no link for author
- [x] Once done on main, will need to be cherrypicked to 2.0 as well
- ~Ensure all people mentioned in any part of attribution is on all-contributors in p5.js README in both main and dev-2.0 branches and include "example" emoji. **Complete list:** `Prof** WM Harris`, `GH/prashantgupta24`~ _Alternatively, can reach out to these contributors and ask them if they want to be included in the list. Since the attribution would be clear on each example._

# Current Behavior

The license is at the bottom of the page, and attribution (inspired by and maintained by) is split.

Screenshot 2025-05-13 at 12 21 45

(Then there is the sketch and the code - viewer has to scroll all the way down)

Screenshot 2025-05-13 at 12 21 51

# PR Behavior

Although attribution of p5.js contributors work is collective, this schema allows detailed attribution of referenced work from outside the community in a detailed attribution history. For example, including the following in the `.mdx` of an example:

```
remix:
  - description: Created by
    attribution:
      - name: Caleb Foss
        URL: https://github.com/calebfoss
      - name: Darren Kessner
        URL: https://github.com/dkessner
    code:
      - label: 2023 code
        URL: https://github.com/processing/p5.js-example/tree/main/examples/01_Shapes_And_Color/01_Color
  - collectivelyAttributedSince: 2024

```

Will create the following attribution:

image

Thus, new examples can be added that correctly attribute referenced material. In the above schema, the `collectivelyAttributedSince` field is applicable mainly to examples ported from the Processing or archived p5.js site. For future additions, the `remix` filed can be used to provide a complete history. The URLs for attribution and type of remix can be edited, with optional URLs. If there is no remixLabel, "Remixed from" will be used, but other phrasing can be provided for clarity.


src/content/ui/en.yaml
```diff
@@ -195,3 +195,14 @@ LibrariesLayout:
 experimentalApi:
   title: This API is experimental
   description: Its behavior may change in a future version of p5.js.
+Remixed by: Remixed by
+Revised in 2023 by: Revised in 2023 by
+Revised by: Revised by
+Based on: Based on
+Created by: Created by
+and: and
```
davepagurek: I think the format you've arrived at here is a good compromise between having to juggle a lot more content in translations and awkwardness in structure.

The one thing that might be an issue, which I'm not really sure is an issue or not, is whether translations for small chunks can be strung together with the same structure across languages, or if the grammar of those languages would need e.g. a reordering of the translation strings. It seems like most of these should be ok, but I guess the code to full sentences we do here, the more flexibility translations will have. This could be a non issue for this current set of strings though, I just don't know enough of the languages we support to say for sure so I thought I'd mention it just in case!

src/layouts/ExampleLayout.astro
```diff
@@ -45,6 +45,22 @@ const relatedReferences =
 
 const { Content } = await example.render();
 
+// Extract the collective attribution year. If multiple provided, uses last shown.
+const collectivelyAttributedSince = example.data.remix?.reduce(
+  (acc: number | null, item) => {
+    if (item.collectivelyAttributedSince) {
+      return item.collectivelyAttributedSince;
```
davepagurek, line 52: I guess this means we need these entries in chronological order? might be worth stating that explicitly in the docs in case someone accidentally does them in reverse chronological.

davepagurek (approved):

src/layouts/ExampleLayout.astro
ksen0: Great point, will update!

src/content/ui/en.yaml
ksen0: Thanks again for taking the time to review this. So after looking into this - the full solution would involve something like a i18n string "Edited and maintained by {p5.js} and {pf_org}" which is then filled in with the link, so more grammars can be supported, without this `and: and` workaround. I will make this a separate open issue for someone else to tackle. In this PR I have updated the i18n strings to se a clearer hierarchy for those strings relevant to the attribution message, so there's no side effects:

```
attribution:
  Remixed by: Remixed by
  Revised in 2023 by: Revised in 2023 by
  Revised by: Revised by
  Based on: Based on
  Created by: Created by
  and: and
  Edited and maintained by: Edited and maintained by
  From 2024 onwards, edited and maintained by: From 2024 onwards, edited and maintained by
  You can find the code history of these examples here: You can find the code history of these examples here
  You can suggest improvements by: You can suggest improvements by
  contributing to the current website: contributing to the current website
```

davepagurek: That approach makes sense, thanks for looking into it!

PR #1029, opened 2025-11-14
Add the every library

mngyuan:
`every` is a library for managing time based events like scene transitions or repetitive tasks. It's partially inspired by `100.times do |n|` from Ruby and helps chain time dependent drawing changes in p5. I've been using it in class with students as a way to ease into time-based media and using time in p5.js outside of looping or `setTimeout`.

There's some more information on how you use it in the `README`:

https://github.com/mngyuan/every/

[Example sketch 1](https://editor.p5js.org/mngyuan/sketches/6hvIBn8b0) [Example sketch 2](https://editor.p5js.org/mngyuan/sketches/9_9gpe_Vh)


davepagurek:
Thanks for adding this, and nice work on the library! Just a heads up, on the website the thumbnails are 3:2 aspect ratio, so a 1500x1000 image would match that. Right now you've got a square 1500x1500, so it will get cropped on the site. Did you want to create a version matching the display aspect ratio?

mngyuan:
Thanks! I'm happy with the way it looks on the site, I've built it locally and took a look. It's designed to be cutoff as it is (or rather it's a happy accident).

ksen0:
Thanks @mngyuan  for building and sharing this!

ksen0:
PS @mngyuan I checked the examples with p5.js 2.0 - out of the box they don't run, but if you add 2.0 support, please make a PR based on `2.0` to add this library there too! Thanks again.

PR #1043, opened 2025-12-03
Add the every library to 2.0

mngyuan:
Edited the every library to work with p5.js 2.0. This PR is the same as #1029 but into the 2.0 branch for the 2.0 website


mngyuan:
the changes to support 2.x can be found [here](https://github.com/mngyuan/every/compare/v1.0.1...v1.1.0)

ksen0:
Thanks for the ping @mngyuan , this is awesome to see!

PR #1052, opened 2025-12-14
Add p5.chart library

tadavoila:


davepagurek (approved):
Thanks for adding this! Does your library also work in p5 2.0? If so, feel free to open a pr with this same content into the `2.0` branch too 🙂

PR #1054, opened 2025-12-22
Add p5.record.js addon library

limzykenneth:
Adding a new addon library p5.record.js that uses native Web APIs to record p5.js sketches.


ksen0:
Thanks, should I cherrypick this into 2.0 as well?

limzykenneth:
@ksen0 Actually it might make more sense to be in 2.0 as it is a 2.x only library.

PR #1070, opened 2026-01-01
Dependencies updates

limzykenneth:
Mainly around the build tools (eg. Astro, Vitest, Preact). Tailwind may be able to be updated as well but will need a bit of time to look into it.

The test may show some error around incompatibility of ESM and CommonJS but for me it is passing anyway so may not be an issue at this point.

I'll do a cherry pick for the main branch after this is merged as well.


ksen0 (approved):

limzykenneth:
@ksen0 I'll cherry pick for the main branch shortly

PR #1072, opened 2026-01-01
Fix #1030: p5.sound reference pages link to correct GitHub repository

### LINKED ISSUE #1030: p5.js 2.x reference pages for p5.sound functions link to wrong repository

Reference

Linux

### Web browser and version

Brave 1.83.112

### Actual Behavior

Reference pages have a link to the source where the reference is embedded at the end. For example, https://beta.p5js.org/reference/p5.sound/p5.oscillator/ has the text "Please feel free to edit src/sources/Oscillator.js and open a pull request!" But for version 2 p5.sound pages, the link points to the p5.js repository (https://github.com/processing/p5.js/blob/v2.1.1/src/sources/Oscillator.js#L10 in this example), which gives a "Page not found" error. This appears to be the case for all p5.sound reference pages.

### Expected Behavior

Links from version 2 p5.sound pages should point to the p5.sound.js repository (https://github.com/processing/p5.sound.js/blob/main/src/sources/Oscillator.js in this example).

### Steps to reproduce

Go to https://beta.p5js.org/reference/p5.sound/p5.oscillator/, scroll to the bottom, and click the src/sources/Oscillator.js link. Observe that it links to a non-existent page.

No

rakesh2OO5:
### Summary
Fixes incorrect GitHub source links on p5.sound reference pages by routing them to the `processing/p5.sound` repository instead of `processing/p5.js`.

### Changes
- Updated reference builder logic to select the correct repository based on module type

### Testing
- Verified generated links point to the correct GitHub repositories

Fixes #1030

### DIFF
--- a/src/scripts/builders/reference.ts
+++ b/src/scripts/builders/reference.ts
@@ -292,7 +292,16 @@ const convertToMDX = async (
 
   try {
     // Add YAML comment to the frontmatter
-    const comment = `# This file was auto-generated. Please do not edit it manually!\n# To make changes, edit the comments in the corresponding source file:\n# https://github.com/processing/p5.js/blob/v${p5Version}/${doc.file.replace(/\\/g, '/')}#L${doc.line}`;
+    const repo =
+  doc.module === "p5.sound"
+    ? "processing/p5.sound"
+    : "processing/p5.js";
+
+const comment =
+  `# This file was auto-generated. Please do not edit it manually!\n` +
+  `# To make changes, edit the comments in the corresponding source file:\n` +
+  `# https://github.com/${repo}/blob/v${p5Version}/${doc.file.replace(/\\/g, '/') }#L${doc.line}`;
+
     
     // Convert the frontmatter to a string
     const frontmatter = matter.stringify("", frontMatterArgs);


davepagurek:
Hi! I think this doesn't quite solve the issue.
- If this is an issue with the 2.x reference, then the PR should be into the `2.0` branch, since this main branch is only for the 1.x site.
- I think this is just editing the comment in the .mdx source files? You're right that this is also likely incorrect and in need of fixing, but the issue is talking not about a code comment, but about a link in the public site. I believe the source for that is in the ReferenceItemLayout file.

rakesh2OO5:
Thanks for the clarification — that helps a lot.

You're right, this issue refers to the 2.x reference site rather than the 1.x site on `main`, and the visible link likely comes from `ReferenceItemLayout` rather than the reference builder comment.

I’ll take a look at the 2.0 branch and trace the source of the public link there. I’m happy to close or repurpose this PR if this change isn’t appropriate here.

Thanks again for pointing me in the right direction.

rakesh2OO5:
Closing this PR — this change targets the reference builder comment on main, but the issue affects the 2.x reference site UI. I’ll follow up with a correct fix on the 2.0 branch.

PR #1074, opened 2026-01-03
Fixed the "edit source" links on p5.sound reference pages (beta.p5js.org) to point to the correct repository.

Linqi1234:
## Summary
Fixed the "edit source" links on p5.sound reference pages (beta.p5js.org) to point to the correct repository.

## Problem
The source links on p5.sound reference pages were pointing to `processing/p5.js`, which resulted in 404 errors. For example:
- ❌ `https://github.com/processing/p5.js/blob/v2.1.2/src/sources/AudioIn.js` (404)

## Solution
Added a conditional check: when `entry.data.module === 'p5.sound'`, the link now points to `processing/p5.sound.js` repository with the `main` branch:
- ✅ `https://github.com/processing/p5.sound.js/blob/main/src/sources/AudioIn.js`

### DIFF
--- a/src/layouts/ReferenceItemLayout.astro
+++ b/src/layouts/ReferenceItemLayout.astro
@@ -272,7 +272,7 @@ const descriptionParts = description.split(
               };
 
               return (
-                <div class="text-body my-lg">
+                  <div class="text-body my-lg">
                   <a
                     href={`/reference/${methodValue.path}`}
                     class="text-body-large"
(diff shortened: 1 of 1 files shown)


ksen0:
Hi, I wll close this as the original problem was resolved by a different fix. Sorry this slipped through. Please note in the [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/) we ask that before making an PR, each volunteer gets assigned on the issue and their approach approved. This helps to avoid accidental duplicate work. I recommend starting with issue in the future . Thanks for understanding!

PR #1078, opened 2026-01-06
Fix search clear button not resetting search state

### LINKED ISSUE #949: Search Cancel button does not delete query text

Other (specify if possible)

Mac OS

### Web browser and version

firefox

### Actual Behavior

After typing a search, then hitting the X to clear the search, it looks like it was cleared, but then when you start typing again the old search text is still there

### Expected Behavior

The X should completely delete the text

### Steps to reproduce

- type in the search field
- hit the X
- start typing again

no thank you

Pratik5252:
Fixes #949 

Previous Behavior
When users clicked the X to clear the search input, the visual field was cleared but the search state in the parent component remained unchanged. This caused old search results to reappear when typing again.

Solution
Updated clearInput() to call onSearchChange("") so the parent component's search state is properly reset along with the input field.

### DIFF
--- a/src/components/SearchResults/index.tsx
+++ b/src/components/SearchResults/index.tsx
@@ -98,6 +98,7 @@ const SearchResults = ({
     if (inputRef.current) {
       inputRef.current.value = "";
     }
+    onSearchChange("")
   };
   const submitInput = () => {
     if (inputRef.current) {


perminder-17 (approved):
Works well and looks good to me! Thanks for the fix

perminder-17:
Are you also interested in fixing that in `2.0` branch? I think the same issue is there with that branch as well.

Pratik5252:
Yes, I'll fix the same in 2.0 branch

Pratik5252:
@perminder-17 I have raised a PR and applied the same changes to branch 2.0

Issue #1085, opened 2026-01-09
unnecessary console output "Component rendered" when embedded sketch is played (CopyCodeButton)


nbogie:
### Most appropriate sections of the p5.js website?

Tutorials

### What is your operating system?

Mac OS

### Web browser and version

chrome 143.0.7499.170 (Official Build) (arm64)

### Actual Behavior

Each time one presses play on an embedded sketch (e.g. [in v1.x reference: fill](https://p5js.org/reference/p5/fill/), or [in beta: strands tutorial](https://beta.p5js.org/tutorials/intro-to-p5-strands/) ) the following text is added twice to the console:

```
Component rendered, isCopied: false
```
This quickly fills up the console with noise as the user iterates on a sketch (e.g. when trying to debug something).

### Expected Behavior

There should be no console output in the happy path when pressing play on an embedded sketch, unless the sketch code itself adds it.

### Steps to reproduce

* open a tutorial (tested with [getting started](https://p5js.org/tutorials/get-started/) and [intro to strands](https://beta.p5js.org/tutorials/intro-to-p5-strands/))
* open dev tools console
* press play on a sketch repeatedly.  
* observe console

### Would you like to work on the issue?

yes, please!

PR #1086, opened 2026-01-09
remove "Component rendered" console output on playing embedded sketch

nbogie:
Removes unnecessary console logging that was occurring each time any embedded sketch was played.

### DIFF
--- a/src/components/CopyCodeButton/index.tsx
+++ b/src/components/CopyCodeButton/index.tsx
@@ -36,8 +36,6 @@ export const CopyCodeButton = ({
     }
   };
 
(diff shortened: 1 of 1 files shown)


nbogie:
fixes #1085

davepagurek:
Thanks!

davepagurek:
We may also need to do the same change in the `2.0` branch, if you're up for making the same change there too.

nbogie:
> We may also need to do the same change in the `2.0` branch, if you're up for making the same change there too.

oops, yeah, #1088 !

PR #1094, opened 2026-01-16
Fix reference sidebar flash by server-rendering directory

### LINKED ISSUE #477: Consider using transitions or reduce use of JS in the nav links
### Increasing Access

This could prevent layout shift and "flashing" on page navigation which could create a more continuous and easier-to-follow experience.

Home

### Feature request details

The nav panel reloads on every navigation. This leads to a kind of "flashing" experience in that component. I know that the related components were switched to [Preact components in order to handle active resizing](https://github.com/processing/p5.js-website/blob/main/src/components/Nav/NavPanels.tsx) from desktop <-> mobile but I think it would be worth the effort to move these back to Astro components with scripts where possible. In my opinion, this would significantly improve look and feel.

An alternative option that could achieve something interesting is [Astro View Transitions](https://docs.astro.build/en/guides/view-transitions/) but those would warrant further testing as they have their own impact on look and feel.

This is in no way urgent but I just wanted to file it to indicate that I think this would be an improvement. I'd be open to working on it at some point but it would likely be a month or so until I had the capacity to take it on

rakesh2OO5:
Fixes #477

### Summary
Server-renders the reference directory to prevent the initial sidebar flash and layout shift caused by client-only hydration.

### Changes
- Added a server-rendered Astro component for the reference directory
- Kept client-side filtering as a progressive enhancement
- Updated ReferenceLayout to render the directory during SSR

### Result
- Sidebar appears immediately on load
- No visible flash or layout jump
- Filtering behavior remains unchanged

### DIFF
--- a/src/components/Nav/NavPanels.tsx
+++ b/src/components/Nav/NavPanels.tsx
@@ -32,16 +32,17 @@ export const NavPanels = (props: NavPanelsProps) => {
     jumpToState,
   } = props;
 
-  const [isOpen, setIsOpen] = useState({ main: false, jump: false });
+  const [isOpen, setIsOpen] = useState({ main: true, jump: true });
   const [isMobile, setIsMobile] = useState(true);
 
-  // Defaults to closed on mobile, open on desktop
-  // Have to do this in a lifecycle method
-  // so that we can still server-side render
   useEffect(() => {
     const startsMobile = window.innerWidth < 768;
     setIsMobile(startsMobile);
     setIsOpen({ main: !startsMobile, jump: !startsMobile });
+    requestAnimationFrame(() => {
+      document.documentElement.setAttribute("data-nav-panels-ready", "true");
+    });
+
     // We use a resize observer to the user's window crosses the
     // threshhold between mobile and desktop
     const documentObserver = new ResizeObserver((entries) => {
--- a/src/components/Nav/styles.module.scss
+++ b/src/components/Nav/styles.module.scss
@@ -132,7 +132,7 @@
 .jumpto {
   border-color: var(--sidebar-type-color);
 
-  animation: height 0.3s ease-in-out;
+  animation: none;
   flex-grow: 0;
   border-bottom-width: 1px;
   height: 40px;
@@ -262,7 +262,9 @@
   margin-top: 4px;
 }
 
-/* Interaction */
+:global(html[data-nav-panels-ready="true"]) .jumpto {
+  animation: height 0.3s ease-in-out;
+}
 
 .jumpto:global(.open) {
   flex-grow: 1;
@@ -277,7 +279,17 @@
     }
   }
 }
+@media (max-width: ($breakpoint-tablet - 1px)) {
+  :global(html:not([data-nav-panels-ready="true"])) .mainlinks:global(.open) {
+    height: 43px;
+  }
 
+  :global(html:not([data-nav-panels-ready="true"])) .jumpto:global(.open) {
+    flex-grow: 0;
+    height: 40px;
+    overflow: hidden;
+  }
+}
 /* Page-specific modifications */
 :global(.reference-item) .jumpto ul {
   grid-template-columns: 1fr;
@@ -286,4 +298,3 @@
 .jumpto ul li.linklabel:not(:global(.small), :first-child) {
   margin-top: 10px;
 }
-


stalgiag:
Thanks! I will try to review this some time in the next month. If any other reviewers gets here first, please by all means feel free to review ahead of me.

coseeian:
I pulled the changes to test them locally, but I noticed some unexpected UX on the page, and it seems like we’re still seeing the flash in the original nav (screenshot below). 

https://github.com/user-attachments/assets/aaa0d032-af4d-460f-b8f4-604a0f21e437

It looks like the 'flashing' persists because the issue might be rooted deeper in how the global NavPanels component handles initial resizing on the client side (the desktop isOpen state being false by default): https://github.com/processing/p5.js-website/blob/main/src/components/Nav/NavPanels.tsx

Since this affects every page and not just the Reference section, I think we might need to shift our focus to that core component to truly solve the flash issue described in https://github.com/processing/p5.js-website/issues/477.

rakesh2OO5:
Thanks for the detailed investigation and for pointing me to NavPanels.tsx.
I’ve converted this PR to draft while I rework the fix to address the root hydration mismatch in the global nav rather than just the reference directory.
I’ll mark it ready for review once the revised approach is pushed.

rakesh2OO5:
Thanks again for pointing me to NavPanels.tsx — that was the key.
I’ve opened a clean follow-up PR that fixes the root hydration mismatch in the global nav with a minimal diff: #1102(closed currently)

I’m leaving this PR in draft as context for the earlier investigation. Happy to close it if you prefer.

davepagurek:
@rakesh2OO5 it is probably easier to follow the discussion if you continue to commit to the same PR rather than opening new ones, as it contains everything to one place.

rakesh2OO5:
That makes sense — thanks for the guidance.
I’ll move the fix into this PR and keep everything here so it’s easier to follow.

rakesh2OO5:
Thanks for the guidance earlier — I’ve now consolidated the fix into this PR with a minimal diff (+36/-26) that addresses the root hydration mismatch in NavPanels and removes the sidebar flash without changing existing behavior.

@coseeian if you have a moment, I’d really appreciate a review on this updated version. Happy to iterate if anything needs adjustment.

src/components/Nav/NavPanels.tsx
```diff
@@ -32,38 +40,40 @@ export const NavPanels = (props: NavPanelsProps) => {
...
-  const [isMobile, setIsMobile] = useState(true);
+  const [isMobile, setIsMobile] = useState(getIsMobile);
+
+const [isOpen, setIsOpen] = useState(() => {
+  const mobile = getIsMobile();
+  return {
+    main: !mobile,
+    jump: !mobile,
+  };
+});
```
coseeian: nit: please fix indention.

src/components/Nav/NavPanels.tsx
```diff
@@ -32,38 +40,40 @@ export const NavPanels = (props: NavPanelsProps) => {
...
+      return mobile;
     });
-    documentObserver.observe(document.body);
-    return () => documentObserver.disconnect();
-  }, [setIsMobile, setIsOpen, isMobile]);
+  };
+
+  window.addEventListener("resize", handleResize);
+  return () => window.removeEventListener("resize", handleResize);
+}, []);
```
coseeian: Nit: Looks like there might be a indentation tweak needed in this code block. Could you take a quick look?

src/components/Nav/NavPanels.tsx
coseeian: Nit: Looks like there might be a indentation tweak needed in this code block. Could you take a quick look?

coseeian (commented):
I pulled this PR down and test locally by navigation to `/reference/p5/arc/`. Noticed regression on mobile. It looks like the changes to address issue on desktop are affecting mobile nav behavior.

Expected: Nav should be closed by default.

Image

Actual: menu nav keep opened, and reference nav closed with flash.

https://github.com/user-attachments/assets/de62fc12-72a4-4368-9cba-e9552b8331df

Please have a look. Thanks.

rakesh2OO5:
Thanks for flagging this — agreed, mobile behavior should stay consistent with the expected default (nav closed).
I’ll adjust the logic to preserve the mobile default state and clean up the indentation in this block. Will push an update shortly.

src/components/Nav/NavPanels.tsx
rakesh2OO5: I will surely indent the code properly and update!

src/components/Nav/NavPanels.tsx
rakesh2OO5: I will surely indent the code properly and update!

rakesh2OO5:
Hi @coseeian — I’ve pushed an updated commit that resolves both the desktop sidebar flash and the mobile nav regression with a minimal diff (+19/-7).

Whenever you have a moment, I’d really appreciate a review. Thanks!

coseeian:
While the root cause lies in the nav component's default state, I’m concerned that toggling it to true simply shifts the "flash" issue from desktop to mobile.

As shown in the recording below, although the issue of nav height is resolved, the hamburger menu button and chevron button in the top right flip states immediately after the page loads. (It's more visible if you slow down the playback speed).

https://github.com/user-attachments/assets/4c91ba65-e320-4d63-95d5-6e87f457f2b2

Since we can't determine device type server-side in this project (correct me if I'm wrong), a client-side React state fix might always result in same issue. An alternative is to refactor and decouple the nav into two device-specific sub-components, each with its own appropriate initial state.

Trade-offs to consider:
* Bundle size: this would introduce more code to the client.
* CSS complexity: we would need to carefully manage the hidden styles via CSS media queries for each component.

Pinging @stalgiag for your thoughts on whether this structural change is worth the overhead.

stalgiag:
That is correct @coseeian. When I originally wrote the issue though, I thought that we could simplify even further by moving away from Preact with dedicated subcomponents that hide/show via CSS (which feels like a workaround) and instead use Astro components. If I remember correctly the only reason for moving to Preact was to handle the default open state (different on mobile vs desktop).

I *think* based on my foggy memory that a small refactor could make it so that the server-rendered HTML picks the right initial CSS classes. Since the server doesn't know the viewport width, we'd want the default HTML to have no open class on either panel, and use a CSS media query to handle the desktop default:                                                   
```                                                                                                                                                    
  @media (min-width: $breakpoint-tablet) {                                                                                                                                
    .main-nav, .jumpto {                                                                                                                                                  
      // desktop default styles (expanded)                                                                                                                                
    }                                                                                                                                                                     
  }                                                                                                                                                                       
```

All of the interactivity could live in a in a small inline script with `matchMedia` and class toggling.

rakesh2OO5:
Thanks for the detailed feedback @coseeian and @stalgiag — that makes sense, and I agree the root issue is the SSR vs hydration mismatch rather than just the client-side default state.

My intent with this PR was mainly to address the visible regression/flash as a short-term fix with a minimal diff, but I agree that a CSS-first + SSR approach (media queries for desktop defaults, no “open” classes in the server HTML, and a small matchMedia/toggle script for interactivity) is the more robust long-term direction and aligns better with the original goal of #477.

If you think it makes sense, I’m happy to:
-----------------------------------------------
-Treat this PR as a stopgap to resolve the current regression, and
-Follow up with a separate PR that refactors the nav toward the Astro/CSS default-state approach you outlined.

Let me know which path you’d prefer and I’m glad to take this forward .

Nwakaego-Ego:
Hi @stalgiag @coseeian, thank you both for the thorough investigation and detailed feedback on this PR. Just following up on @rakesh2OO5's question from January about whether to treat this as a stopgap fix or proceed with the full CSS-first refactor you outlined. Could you give them direction so they can move forward?  Thank you, @rakesh2OO5, for your work on this so far.

PR #1098, opened 2026-01-20
fix: clarify frameCount usage in Variables and Change tutorial

### LINKED ISSUE #1097: Confusing explanation of frameCount % 290 and cloudOneX in Variables and Change tutorial

Tutorials

None

### Web browser and version

_No response_

### Actual Behavior

In the Variables and Change tutorial, the explanation incorrectly states that `frameCount % 290` is the value stored in the variable `cloudOneX`.

However, in the code:
- `cloudOneX` is assigned `frameCount % width`
- `frameCount % 290` is used directly in the tree animation logic

This creates a mismatch between the explanation and the actual code behavior.

### Expected Behavior

The explanation should accurately describe which values are stored in which variables.

`frameCount % 290` should not be described as the value stored in `cloudOneX` when that is not the case in the code.

### Steps to reproduce

1. Open the Variables and Change tutorial.
2. Read the explanation describing the use of `frameCount % 290` and `cloudOneX`.
3. Compare the explanation with the corresponding code in the tutorial.

Yes, I can work on this issue and submit a fix if that's ok.

aashu2006:
### Summary
This PR fixes a confusing explanation in the Variables and Change tutorial that incorrectly associates `frameCount % 290` with the variable `cloudOneX`.

### Changes
- Removed an incorrect reference to `cloudOneX` so the explanation accurately reflects the code.

### Related Issue
Resolves #1097


aashu2006:
This change only removes a reference to an internal variable name to keep the tutorial text consistent. Happy to adjust if needed. Thanks!

PR #1100, opened 2026-01-20
fix typography 2.0 tutorial crash caused by space character passed in textToModel()

### LINKED ISSUE #1099: Typography 2.0 tutorial crashes when text contains spaces
The Typography 2.0 tutorial crashes when the displayed text contains space characters. (https://beta.p5js.org/tutorials/typography-20/)

The tutorial code calls `textToModel()` on each character, but space characters produce no geometry and currently throw an error.

Example string from the tutorial:

"P5.JS 2.0"

Cause:

`textToModel(' ')` returns undefined and later rendering code expects
a valid model.

skyash-dev:
This updates the greeting text in the Typography 2.0 tutorial to avoid calling `textToModel()` on a space character, which currently throws due to a regression.

This is a temporary workaround until the core [issue](https://github.com/processing/p5.js/issues/8428) is fixed in p5.js.

Closes #1099


davepagurek:
Thanks for the quick fix!

PR #1107, opened 2026-01-26
fix: restore focus after search submission (#914)

### LINKED ISSUE #914: [focus] Focus is lost after clicking “Submit search” button
### Title

[focus] Focus is lost after clicking “Submit search” button

### Description

When a user clicks the “Submit search” button, the button is removed from the DOM and no programmatic focus management is applied. As a result, keyboard and screen reader users lose track of their place in the interface, which creates confusion and interrupts their workflow. This issue presents a barrier for users who rely on assistive technologies or keyboard navigation, as they cannot easily determine the next actionable element or understand the result of their action.

### Steps to Reproduce

1. Navigate to `/search/`
2. Locate the “Search through site content” input field.
3. Enter a search term into the input.
4. Move focus on the “Submit search” button and click it.
5. Observe that the “Submit search” button disappears and a “Clear search input” button appears.
6. Note that focus is not placed on any element.

### Actual Behavior

After the “Submit search” button is click and removed, focus is lost and resets to the browser’s default navigation starting point (e.g., the document `<body>` or address bar).

### Expected Behavior

When the “Submit search” button is click and replaced, focus should be programmatically moved to the logical appropriate interactive element—such as the “Clear search input” button or returned to the “Search through site content” input field—so that users can seamlessly continue their interaction.

### Environments

_No response_

### Suggested Fix

Implement focus management logic so that after the “Submit search” button is clicked and removed:

* **Option 1:** Move focus to the “Clear search input” button that appears.
* **Option 2:** Return focus to the “Search through site content” input field.

### Reference

* [WCAG 2.2:  Success Criterion 2.4.3 Focus Order](https://www.w3.org/TR/WCAG22/#focus-order)
* [ARIA Authoring Practices Guide: Managing Focus](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/)

None

### Web browser and version

_No response_

coseeian:
## Summary

Fixes #914 - Restores focus management after search submission to improve keyboard navigation and screen reader accessibility.

**Problem:** When users clicked the "Submit search" button, focus was lost as the button disappeared from the DOM and was replaced with the "Clear search input" button. This created an accessibility barrier for keyboard and screen reader users who lost their place in the interface.

**Solution:** Implemented automatic focus transfer to the clear button after search submission using `useEffect` to track state transitions. Also removed the `key` prop that was causing component remounts so that we can preserve the state tracking required for focus management.

**Changes:**
  - Add `clearButtonRef` and focus management logic to automatically focus clear button after search submission.
  - Track previous `isInputEdited` state to detect transitions (prevents unwanted focus on initial render).
  - Remove `key={searchTerm}` prop from SearchResults - this was causing the component to unmount/remount on every search, which would clear the `prevIsInputEdited` state and break focus management. Replace remount-based state reset with `useEffect` for better performance and to preserve component state.

## Demo

After

https://github.com/user-attachments/assets/469b0ea3-dd25-4eea-8729-79f8b0881142

### DIFF
--- a/src/components/SearchProvider/index.tsx
+++ b/src/components/SearchProvider/index.tsx
@@ -126,7 +126,6 @@ const SearchProvider = ({
       currentLocale={currentLocale as string}
       onSearchChange={handleSearchTermChange}
       uiTranslations={uiTranslations}
-      key={searchTerm}
     />
   );
 };
--- a/src/components/SearchResults/index.tsx
+++ b/src/components/SearchResults/index.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useRef, useState } from "preact/hooks";
+import { useEffect, useMemo, useRef, useState } from "preact/hooks";
 import { Icon } from "../Icon";
 
 type SearchResult = {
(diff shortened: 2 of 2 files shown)


perminder-17 (approved):
Looks good to me.

ksen0:
Awesome, thank you!

PR #1121, opened 2026-01-29
Fix broken p5.* reference links in reference pages

### LINKED ISSUE #1120: Links to `p5.Element` don't work correctly

Reference

Mac OS

### Web browser and version

Brave: 1.85.120 Chromium: 143.0.7499.192 (Official Build) (arm64) 

### Actual Behavior

Each page for a component that interacts with `p5.Element`, such as `createButton`, `createSlider`, `createDiv`,  etc, and even `createElement` all have a bad link to the `Element` reference page, instead staying on current page. 

This is located in the "Returns" section of the page, as each of these components return a `p5.Element` type object.

Each of these pages experiences the same issue of a malformed URL.

Create div has this URL: `https://p5js.org/reference/p5/createDiv/`
The link to the `p5.Element` is this: `https://p5js.org/reference/p5/createDiv/#/p5.Element` which does nothing but scroll elsewhere in the same page. 

This is particularly frustrating while teaching in a classroom setting. The easiest way to get to the Element page is to search for "element" in the top bar and select "p5.Element" from the results which is unnecessary cognitive load and a huge interruption to flow of thought. **The p5.Element page is necessary for all of these as that is where the method reference is location.**

### Expected Behavior

Under the "Returns" section of each of these pages the link to `p5.Element` should be `https://p5js.org/reference/p5/p5.Element/`

### Steps to reproduce

1. Go to a reference page of a component that returns a `p5.Element`, which would be almost everything in the DOM section.
2. Scroll to "Returns"
3. Click the link in this line:

> p5.Element: new [p5.Element](https://p5js.org/reference/p5/createDiv/#/p5.Element) object.

Example (same as bove):

Page:`https://p5js.org/reference/p5/createDiv/`
The link to the `p5.Element` is this: `https://p5js.org/reference/p5/createDiv/#/p5.Element` which does nothing but scroll elsewhere in the same page. 

sure but not really

rakesh2OO5:
This PR normalizes malformed #/p5.* links in reference pages by rewriting them to /reference/p5/{type}/ at render time.

The fix applies to both Returns and Parameter descriptions, ensuring p5.Element, p5.Vector, and similar types route correctly across all DOM reference pages.

Tested locally on createDiv() .
Closes #1120.


ksen0:
Thanks for the patience. Will merge. Tested with:

* Original problem: Create div has this URL: https://p5js.org/reference/p5/createDiv/
The link to the p5.Element is this: https://p5js.org/reference/p5/createDiv/#/p5.Element which does nothing but scroll elsewhere in the same page.
* Sound part of the references: http://localhost:4321/reference/p5.Envelope and http://localhost:4321/reference/p5.Envelope/releaseTime/

PR #1170, opened 2026-02-14
format 2.0 typography tutorial for better readability

### LINKED ISSUE #1169: 2.0 typography tutorial has indentation and formatting problems

Tutorials

Mac OS

### Web browser and version

145.0.7632.45 (Official Build) (arm64)

### Actual Behavior

As shown in the image, 
* there's no indentation of the body of setup or draw.  This is the case throughout, for any code block using `<EditableSketch/>`
* the unintentionally long banner comments wrap, distracting

Also, not shown: 
* various lines wrap awkwardly, making the code harder to follow

Image

Image

Image

### Expected Behavior

As far as possible, lines should not wrap. Where they do they should not distract from the code.

### Steps to reproduce

1. visit https://beta.p5js.org/tutorials/typography-20/ 
2. make screen window as wide as possible
3. visit on mobile, too

yes!

nbogie:
Fixes #1169

### changes
Edited only [src/content/tutorials/en/typography-2.0.mdx](src/content/tutorials/en/typography-2.0.mdx)

1. For all uses of EditableSketch, add two spaces at the start of every line in the code span.  Markdown parser seems to eat two from each line that has them (before the component gets access).
  This way, the remaining _relative_ indentation the author sees is what will be rendered.
  I would suggest it for all tutorials using EditableSketch or code spans.
  Code fences with triple-backtick do not have this problem.

2. Wrap various long code lines intentionally, for readability
3. make banner comments shorter.

### testing

Tested all examples are still running
Manually checked wrapping and readability on the tutorial
  * at 765px width (just under 770px breakpoint where the side banner gets hidden)
  * at 1920px width

### screenshots

* Bodies of setup, draw functions are now indented.  
* Banner comments don't wrap (at full width window).

image

image

### DIFF
--- a/src/content/tutorials/en/typography-2.0.mdx
+++ b/src/content/tutorials/en/typography-2.0.mdx
@@ -39,22 +39,22 @@ We'll use some of p5.js 2.0's 3D features to do our block letters, so we'll crea
 When debugging a 3D sketch, adding `orbitControl()` lets us click and drag to rotate the things we've drawn. If you run the sketch now, you'll see a box that you can spin around by clicking and dragging.
 
 <EditableSketch code={`
-function setup() {
-  /////////////////////////////////////////////////////
-  // Switch to WEBGL mode at postcard size
-  /////////////////////////////////////////////////////
-  createCanvas(600, 400, WEBGL);
-}
-
-function draw() {
-  background(255);
+  function setup() {
+    /////////////////////////////////////////////
+    // Switch to WEBGL mode at postcard size
+    /////////////////////////////////////////////
+    createCanvas(600, 400, WEBGL);
+  }
   
-  /////////////////////////////////////////////////////
-  // Draw a box that you can orbit around
-  /////////////////////////////////////////////////////
-  orbitControl();
-  box(50);
-}
+  function draw() {
+    background(255);
+    
+    /////////////////////////////////////////////
+    // Draw a box that you can orbit around
+    /////////////////////////////////////////////
+    orbitControl();
+    box(50);
+  }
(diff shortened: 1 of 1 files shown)


src/content/tutorials/en/typography-2.0.mdx
```diff
@@ -1016,7 +1068,7 @@ function draw() {
   image(
     paperImg,
     0, 0, width, height, // Destination size
-    0, 0, paperImg.width, paperImg.height, // Source size
+    0, 0, paperImg.width, paperImg.height, //Src size
```
ksen0: Very minor: I think `// Source size` is better (more direct, the space is good?)

ksen0 (approved):
Just 1 minor note @nbogie , thanks for this! Happy to merge or another maintainer can.

src/content/tutorials/en/typography-2.0.mdx
nbogie: Thanks for reviewing!  

This one was a compromise.  I'd absolutely agree "Source size" is better but that it makes the line wrap and the code harder to read as a result.  I think `src` is _very_ well known abbreviation for source even for newcomers, given its use in `` and `<script src="" />` even if they haven't met a `src/` folder yet.  (It is also used in param names in two p5.Image methods).

Other options: 
* `// Source` instead of `// Src size`.
* removing the comment later when the line lengths increase

Both of these I think would be fine as they occur in a pair with `// Destination size` something like 15 times throughout the tutorial!

I think i'll change to `// Source` assuming (iirc) it fits.

Issue #1187, opened 2026-02-20
Build fails with getCurationSketches when OP API inaccessible


ksen0:
### Most appropriate sections of the p5.js website?

Other (specify if possible)

### What is your operating system?

Other (specify if possible)

### Actual Behavior

Build is failing for multiple PRs because of `getCurationSketches` break. That code contains pollng the OpenProcessing API.

<img width="1166" height="418" alt="Image" src="https://github.com/user-attachments/assets/a58139de-ac13-44cd-b5f9-cb0f7e686c35" />

<img width="1167" height="260" alt="Image" src="https://github.com/user-attachments/assets/f8e63b93-df2c-42a2-a0ae-10e5067d409c" />

### Expected Behavior

Should be possible to fetch OP. Should have a fallback: if OP fetched data already exists it should not be re-fetched at every new CI run. 

### Steps to reproduce

See failed builds for:
* https://github.com/processing/p5.js-website/actions/runs/22182341582/job/64146661413
* https://github.com/processing/p5.js-website/pull/1183
* https://github.com/processing/p5.js-website/pull/1184
* https://github.com/processing/p5.js-website/pull/1186

### Would you like to work on the issue?

I can but if someone else wants to help investigate and fix they are welcome!

kajal-jotwani:
Hi @ksen0!
I checked the failing CI logs and it looks like `getCurationSketches` throws when OP returns 500/429, which then breaks the static build.
I think we could add a persistent fallback for example, saving the fetched OP data to a JSON file after a successful request and reusing it in future CI runs instead of re-fetching each time, along with handling API failures more gracefully.
If that sounds okay, can I work on this issue?

priyaa-j:
Hi @ksen0 and @kajal-jotwani! I encountered this exact getCurationSketches 403/500 error while working on the broken asset cleanup in #1189. I've been using a local bypass to keep my environment functional, but I’d love to assist in implementing a permanent fix.
@kajal-jotwani, I'm happy to help test your fallback logic or assist with the JSON data structure to ensure it handles these API failures gracefully. I would love to help verify the fix on my end or assist in any way. Thanks!

ksen0:
Thank you both @kajal-jotwani @priyaa-j ! Since you commented first @kajal-jotwani I'll assign you. But thanks a lot for your input @priyaa-j I hope you also support with review.

ksen0:
[...]

PR #1195, opened 2026-02-23
Build from cached JSON files

ksen0:
Addresses https://github.com/processing/p5.js-website/issues/1187

This is a bypass. A separate PR to contain the script for generating this cache, but this is standalone. For visibility, script is here: https://gist.github.com/ksen0/6c1c1c592db57b530c63c02089d8e31b Full disclosure: used copilot for some of this

Note that the real solution may be to move to client-side calls; however, need to verify that this would not cause additional/new rate limit issues. In the meantime, builds must be unblocked.


ksen0:
Based on @limzykenneth 's input on Discord, will merge this as a stopgap with strong intent to remove checked in files as soon as possible

PR #1201, opened 2026-02-25
fix: resolve visualID render crash in Sketch component

### LINKED ISSUE #1191: [BUG] Critical rendering crash in Sketch.astro when OpenProcessing API fails

Home

Windows

### Web browser and version

_No response_

### Actual Behavior

Currently, the `src/components/GridItem/Sketch.astro` component assumes that the item prop passed from the OpenProcessing API will always be valid. However, if the API returns an error (such as the recent 403 Forbidden or 500 Internal Server Error), the item variable becomes undefined.
When the component attempts to execute the following line:
`const imgSrc = await getSketchThumbnailSource(item.visualID)`
It throws an `Uncaught TypeError: Cannot read properties of undefined (reading 'visualID').` In the Astro rendering engine, this results in a critical failure that can prevent the entire page (or the sketch grid section) from rendering.

### Expected Behavior

The website should be resilient to external API failures. If a specific sketch's data is missing, the component should exit gracefully (render nothing) instead of crashing the entire application.

Proposed Fix (Guard Clause):
Implement a defensive guard clause at the top of the `src/components/GridItem/Sketch.astro` frontmatter.

### Steps to reproduce

1. Run the server locally.
2. Bypass API Error: Manually bypass/ignore the `getCurationSketches` 403/500 error (#1187) to allow the build to proceed.
3. Observe Error: Even with the API error bypassed, the page fails to render with:
`TypeError: Cannot read properties of undefined (reading 'visualID') at src/components/GridItem/Sketch.astro.`

Yes

priyaa-j:
Closes #1191 
Description:
This PR adds a guard clause in `Sketch.astro`. Currently, when the OpenProcessing API fails (like a 403 or 500 error), the item variable becomes undefined, and the code tries to access `item.visualID`, which causes a `TypeError `and crashes the entire page. 

Changes:
The fix adds a simple check at the top: `if (!item) return` This means if item is undefined or null, the component stops rendering immediately and never tries to access `item.visualID` so the page loads safely without crashing. 

Verification:
The issue was already happening when the API failed when I run the server locally.

**Without fix:** API fails → `item` undefined → page crashes ❌  
**With fix:** API fails → guard clause stops execution → page loads safely ✅

Tested locally on real API failure. Works perfectly.

### DIFF
--- a/src/components/GridItem/Sketch.astro
+++ b/src/components/GridItem/Sketch.astro
@@ -13,6 +13,7 @@ interface Props {
 }
 
 const { item, lazyLoad, scale = 1 } = Astro.props;
+if(!item) return
 const imgSrc = await getSketchThumbnailSource(item.visualID)
 const width = Math.min(thumbnailDimensions * scale, 1500);
 const height = Math.min(


ksen0:
Closing for now, upstream bug was resolved - see linked issue

Issue #1204, opened 2026-02-26
The examples on the createCapture reference don't display well


davepagurek:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Mac OS

### Web browser and version

Firefox

### Actual Behavior

If you go to https://beta.p5js.org/reference/p5/createcapture/, you will see:

- The first example just shows the top corner of the video:
  <img width="1029" height="297" alt="Image" src="https://github.com/user-attachments/assets/3b0a46ed-08d2-4723-a4e5-b4264ae811e9" />

- The third and fourth examples show nothing (especially if, unlike me, you use the Mac default behaviour and scrollbars auto hide) until you scroll
  <img width="1047" height="797" alt="Image" src="https://github.com/user-attachments/assets/ddeb1cde-4af0-49db-a8ff-8e6d81a8960a" />

### Expected Behavior

Ideally you can see the full webcam feed immediately.

So this is happening because we auto size examples based on the canvas, but many of these don't use any canvas, or have a canvas but don't use it so it's just sitting there empty taking up space. While we probably need to update the examples a bit in the p5.js repo, we likely also would need something on the website repo to be able to work too. Maybe we add a new custom tag to an example jsdoc so that we can manually specify a width/height when we know auto detection will fail?

### Steps to reproduce

Go to https://beta.p5js.org/reference/p5/createcapture/

### Would you like to work on the issue?

Feel free for someone else to take this on!

aashu2006:
I would be happy to work on this. It feels like the issue comes from auto sizing based only on canvas detection, which does not work well for DOM based examples like `createCapture()`. I can explore adding a way to manually specify example dimensions (maybe via a custom jsdoc tag) and see how it fits into the reference site rendering logic.

suruanush:
I can make it work and I can fix it.

davepagurek:
Thanks @aashu2006, I'll assign this to you. Let me know what you come up with!

@suruanush also feel free to suggest ideas and offer feedback on @aashu2006's ideas and code.

aashu2006:
Hi @davepagurek , it looks like the preview sizing is based only on canvas detection. For cases like `createCapture()` (with `noCanvas()` or larger DOM elements like `<video>`), the container ends up too small, which causes the clipping or scroll issue.

I'm thinking we could start by adding a sensible fallback size when no usable canvas is detected, before introducing something like a custom `jsdoc` tag. Let me know if that direction makes sense then I’ll proceed.

[...]

PR #1218, opened 2026-03-04
Add fallback preview size for DOM-only reference examples

aashu2006:
Addresses #1204

Example previews on the reference site currently rely on detecting canvas dimensions via `createCanvas()`.

For DOM-only examples (e.g. `createCapture()` used with `noCanvas()`), no canvas dimensions are detected, causing the preview container to collapse or clip the output.

This PR adds a fallback preview size (400×300) when both `previewWidth` and `previewHeight` remain undefined. This ensures DOM-based examples render a visible preview.

### Changes
- Add fallback preview size (400×300) when no canvas dimensions are detected
- Keep existing behavior for sketches that use `createCanvas()`
- Place fallback after the DOM heuristic and before the `largeSketch` calculation

### DIFF
--- a/src/components/CodeEmbed/index.jsx
+++ b/src/components/CodeEmbed/index.jsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useRef } from "preact/hooks";
-import { useLiveRegion } from "../hooks/useLiveRegion";
+import { useLiveRegion } from '../hooks/useLiveRegion';
 import CodeMirror, { EditorView } from "@uiw/react-codemirror";
 import { javascript } from "@codemirror/lang-javascript";
 import { cdnLibraryUrl, cdnSoundUrl } from "@/src/globals/globals";
@@ -38,19 +38,14 @@ export const CodeEmbed = (props) => {
   );
 
   let { previewWidth, previewHeight } = props;
-  const canvasMatch =
-    /createCanvas\(\s*(\d+),\s*(\d+)\s*(?:,\s*(?:P2D|WEBGL)\s*)?\)/m.exec(
-      initialCode,
-    );
+  const canvasMatch = /createCanvas\(\s*(\d+),\s*(\d+)\s*(?:,\s*(?:P2D|WEBGL)\s*)?\)/m.exec(initialCode);
   if (canvasMatch) {
     previewWidth = previewWidth || parseFloat(canvasMatch[1]);
     previewHeight = previewHeight || parseFloat(canvasMatch[2]);
   }
 
   // Quick hack to make room for DOM that gets added below the canvas by default
-  const domMatch = /create(Button|Select|P|Div|Input|ColorPicker)/.exec(
-    initialCode,
-  );
+  const domMatch = /create(Button|Select|P|Div|Input|ColorPicker)/.exec(initialCode);
   if (domMatch && previewHeight) {
     previewHeight += 100;
   }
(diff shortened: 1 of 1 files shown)


aashu2006:
Hi @ksen0 @davepagurek 

As discussed this PR adds a fallback preview size when no canvas dimensions are detected.  
DOM-only examples like `createCapture()` with `noCanvas()` were collapsing because `previewWidth` and `previewHeight` stayed undefined.

The change introduces a 400×300 fallback so those examples render a visible preview, while sketches that use `createCanvas()` continue to use their detected size.

Let me know if this direction looks good or if you would prefer a different fallback approach.

src/components/CodeEmbed/index.jsx
```diff
@@ -38,20 +38,34 @@ export const CodeEmbed = (props) => {
...
+  );
   if (domMatch && previewHeight) {
     previewHeight += 100;
   }
 
+  // Fallback preview size when no usable canvas dimensions are detected.
+  // Ensures DOM-based examples (e.g., createCapture with noCanvas) are visible.
+  const DEFAULT_PREVIEW_WIDTH = 400;
+  const DEFAULT_PREVIEW_HEIGHT = 300;
+  if (previewWidth === undefined && previewHeight === undefined) {
```
ksen0: Edge case, but if only one one is defined correctly, this block is not used, and on line 67 one of them is undefined. `||` probably better?

src/components/CodeEmbed/index.jsx
```diff
@@ -86,7 +100,7 @@ export const CodeEmbed = (props) => {
     >
       {props.previewable ? (
         <div
-          className={`ml-0 flex w-fit gap-[20px] ${largeSketch ? "flex-col" : (props.allowSideBySide ? "" : "flex-col lg:flex-row")}`}
+          className={`ml-0 flex w-fit gap-[20px] ${largeSketch ? "flex-col" : props.allowSideBySide ? "" : "flex-col lg:flex-row"}`}
```
ksen0: Even if this doesn't change behavior, for legibility, I think the () should be reinstated

ksen0 (changes requested):
Thanks! Just 2 small things. Also, if possible, could you please revert unrelated changes in this and future PRs? It's harder to review when there's additional spacing/etc lines changed, so it would be really nice to exclude these. With things like regex especially having a more reliable gitblame is also good

aashu2006:
Thanks for the review @ksen0 💜
The unrelated formatting changes happened due to the formatter running locally when I saved the file. I’ve now reverted those so the diff only contains the relevant logic changes. Let me know if everything looks good now or if anything else should be adjusted.

src/components/CodeEmbed/index.jsx
aashu2006: resolved !

src/components/CodeEmbed/index.jsx
aashu2006: Yes, I have updated it now.

ksen0 (approved):
I'll merge this one to fix the DOM-only examples, but I've changed the PR as "Addresses" because the other examples are still not showing up right (unrelated to size)

aashu2006:
Yes, this PR only targets the DOM only cases where no canvas size is detected which was causing the preview container to collapse. The remaining examples already define a canvas size, so their rendering issue seems to be coming from a different part of the preview pipeline rather than the sizing logic. I am still figuring it out, will update further.

PR #1219, opened 2026-03-05
feat: Add Double Pendulum physics example with Verlet Integration

bryanhoffman:
This PR introduces a new educational example for the "Math and Physics" section. It demonstrates a Double Pendulum using Verlet Integration, offering a more stable and physically accurate alternative to standard Euler-based simulations.

Technical Implementation:

Integration: Uses Position-Based Dynamics (Verlet) for increased numerical stability.

Constraint Solver: Implements a Gauss-Seidel relaxation loop (15 iterations) to maintain rigid rod lengths.

Energy Management: Includes a stochastic Hamiltonian preservation factor. This counteracts numerical damping and simulates a "driven" system (similar to a pendulum on a rickety train), ensuring the system explores its full phase space without losing kinetic energy.


bryanhoffman:
I wanted to try and create an example for the p5.js website. I hope this example is worthy of publishing, but if it is not, I'd love the feedback to get it there.

ksen0:
Hi @bryanhoffman ! There's no open call for examples right now so it will likely take some days to have capacity to review this properly, but since you closed it that means you don't want this considered after all?

If you'd like to suggest this example, I would recommend making an issue where you link to this example and propose it as an addition so others can more easily comment. ([Contributor Guidelines](https://p5js.org/contribute/contributor_guidelines/)describe the process: issue/discussion first, then approval/assignment, then work)

Also keep in mind that translations should not be AI-generated. There's a general [AI usage policy](https://github.com/processing/p5.js/blob/main/AI_USAGE_POLICY.md) too but in this case just be sure to only suggest translations for languages you're fluent in. For every language, at least one steward in that language will have to review.

bryanhoffman:
I'm not sure how I closed it. That was an error. I'm still very new to contributing on GitHub. Indeed all of the translations were done with AI. I should've taken a better look at the Contributor Guidelines. I'll defer to you @ksen0. If you think the example merits consideration, I'll go through the process as described in the Contributor Guidelines. More than anything I wanted the practice and this was good for me: I'll be studying physics soon and I have tons of math to review before I start so this was some great practice.

PR #1223, opened 2026-03-09
feat: added the logic for generating reference docs from  local custom build of p5.sound.js

### LINKED ISSUE #1224: feat: no way to test local docs build of p5.sound.js on the reference page
### Increasing Access

It would allow future contributors to test the local docs build of p5.sound.js on the reference page.

p5.js-website docs parsing script

### Feature request details

while working on p5.sound.js codebase I saw no option to test the generated docs locally from a local build of p5.sound.js although dummy console.log was present but no logic.
https://github.com/processing/p5.js-website/blob/580ec4fb7781b794ce49a2d49c380ea9f7b3735d/src/scripts/parsers/reference.ts#L55-L57

the logic is present for p5.js builds, we can test local builds of p5.js.
https://github.com/processing/p5.js-website/blob/580ec4fb7781b794ce49a2d49c380ea9f7b3735d/src/scripts/parsers/reference.ts#L43-L49

it would be a good idea to implement this so in future we could seamlessly test any doc related changes of p5.sound.js on this website.

this targets the 2.0 branch.

kunstewi:
resolves #1224 

while working on p5.sound.js codebase I saw no option to test the generated docs locally from a local build of p5.sound.js although dummy console.log was present but no logic.
https://github.com/processing/p5.js-website/blob/580ec4fb7781b794ce49a2d49c380ea9f7b3735d/src/scripts/parsers/reference.ts#L55-L57

the logic is present for p5.js builds, we can test local builds of p5.js.
https://github.com/processing/p5.js-website/blob/580ec4fb7781b794ce49a2d49c380ea9f7b3735d/src/scripts/parsers/reference.ts#L43-L49

so i implemented the p5.sound.js local build testing the same way it's for p5.js by specifying the p5.sound.js local build path at LOCAL_P5_SOUND_PATH variable in .env file.

updated the docs on how to use the p5.sound.js local build.

extended the cleanup logic to identify p5.sound.js builds

@ksen0 @limzykenneth please take a look whenever you have time.

Thanks a lot for your time.


kunstewi:
The p5.sound.js going to need active documentation updates as more methods are added and as previous methods are properly documented. Without this there's no way to test p5.sound.js builds on the reference website locally.

src/scripts/parsers/reference.ts
```diff
@@ -61,12 +61,21 @@ export const parseLibraryReference =
...
-        'https://github.com/processing/p5.sound.js.git',
-        'main'
-      );
+      // Clone or copy p5.sound.js
+      if (process.env.LOCAL_P5_SOUND_PATH) {
+        console.log(`Copying local p5.sound.js from ${process.env.LOCAL_P5_SOUND_PATH}`);
+        await fs.cp(process.env.LOCAL_P5_SOUND_PATH, localSoundPath, {
+          recursive: true,
+          // Ignore node_modules and .git to speed up the copy
+          filter: (src) => !src.includes('node_modules') && !src.includes('.git')
```
nbogie: This filter is a bit wide - it will silently reject any file with either of those substrings in its path, e.g. `cante.gitano.js`.  Probably not very likely but would likely waste someone's time on a mystery, if it occurred.

nbogie (commented):
Looks ok! and tests out ok on happy path.
There's no cleanup if fails before the end of the build, I think.

nbogie:
I'm a little curious why you didn't match the pattern the website build uses for parameterisable p5 source: P5_REPO_URL (works with local file path AND actual urls) and P5_BRANCH.

However, while that _would_ be more consistent, I can see it'd maybe also be overkill for what is needed when working with p5 sound.  What you've done seems to fullfill what's described in the issue just fine!

ksen0:
Thanks for the review @nbogie ! @kunstewi no rush of course and thanks for the patience on this, will you have capacity to finish this one up? Cleanup on failure would be good to add, as would a more precise filter, if its straightforward enough.

nbogie:
Because this PR is taking an approach of directory-copy rather than using git clone, it's worth noting that unintended copies of other files and directories from the local p5.sound are currently likely to happen:

### Examples:
* .env - secrets! (i don't know why we'd have secrets in a local p5.sound but it's not impossible.  during CI experiments, for example)
* .claude/ or .codex/ etc - AI agent settings, possibly with API keys (e.g. for MCP servers)
* .worktrees/ (increasingly commonly used for working on parallel tasks with git - duplicate copies of files)
* .bun/, .pnpm-store/ - other tooling some people might be using

The current filter only prevents copying .git and node_modules.

### Suggestion:
How about not copying across _anything_ that starts with a period, except .github (if required).  (and of course continuing to omit node_modules, too).

kunstewi:
@nbogie understood, already working on it.
 
 Thanks a lot for your valuable review.

src/scripts/parsers/reference.ts
kunstewi: updated the filter to include .github and discard other hidden dotfiles, we can update the filter as needed later by adding keywords if needed.

ksen0:
Thank you so much for all your work @kunstewi and review @nbogie ! @ogbabydiesal and I were able to use this locally and it's really great to have it be available. Thanks especially for your patience, since this really took so long! Merging it in when the checks pass.

kunstewi:
@ksen0 I designed it initially as a very quick local feature for getting my work done to test out p5.sound.js docs. But after reading nbogie's review comments, i was working on a rewrite that would match the build structure used to generate local p5.js docs with failure cleanup. But since it's merged should i drop the rewrite or continue? 

I am sorry for taking way longer than it should to do these tasks.

ksen0:
@kunstewi definitely no need to apologize for taking the time, thanks for being persistent with it. Please do finish the rewrite and submit a PR when you can!

The suggested improvements are very sensible but in my opinion not blocking. I don't foresee this being used widely right now so it's ok if it's a little finicky, but it's important for unblocking p5.sound.js work and useful in the current stage. But it will be great for this to improved.

When you make a new PR please do link to this conversation for context so other website stewards know why the PR is there. Please also @ me in the new PR so I don't miss it

Thanks for all your work!

kunstewi:
Understood. Thanks a lot. 😊

PR #1225, opened 2026-03-10
fix: replaced the old script that was blocking p5.sound.js multiple files' docs generation

### LINKED ISSUE #1226: bug: script in reference.ts is buggy resulting in failure of parsing docs of many methods in p5.sound.js

Reference

Mac OS

### Web browser and version

Helium: 0.9.4.1 (Official Build) (arm64) 

### Actual Behavior

check out this issue by dave to see the problem in p5.sound.js - [here](https://github.com/processing/p5.sound.js/issues/72)

When i decided to dig deeper the first thing I found that there was no way to test p5.sound.js reference docs locally on p5.js-website.

so the first thing I did was to update the p5.js-website script to support p5.sound.js docs local testing. I have raised a PR for this - [here](https://github.com/processing/p5.js-website/pull/1223). It would be needed in the future testing p5.sound.js docs.

After this I tried updating class names, method tag, for tag every possible combination but the changes were not reflecting on the p5.js-website reference. I was testing on beta.p5js.org for information as p5.sound.js is on 2.0 beta website.

Then after diving deeper i finally found the problem which was this old script in reference.ts file:

```js
      // Fix p5.sound classes
      for (const key in soundData.classes) {
        const newName = `p5.${  soundData.classes[key].name}`;
        const updated = {
          ...soundData.classes[key],
          name: newName,
        };
        soundData.classes[newName] = updated;
        delete soundData.classes[key];
      }
      for (const item of soundData.classitems) {
        item.class = `p5.${  item.class}`;
      }
```

The reason those functions did not show up on the reference website is due to a small bug in how the parser grouped p5.sound's global functions.

In the JSDoc comments for p5.sound.js, global functions like userStartAudio, getAudioContext, and loadSound state that they belong to the p5.sound module. This led the parser to temporarily assign their class to "p5.sound". Yet, the build script in reference.ts had a simple loop that added "p5." in front of every parsed class name. Because of this, "p5.sound" was changed to "p5.p5.sound".

The Astro website organizes the reference documentation by class. Since there is no actual class called p5.p5.sound, the generator placed these Markdown files into a "black hole" folder (src/content/reference/en/p5.p5.sound/) and they were cut off from the sidebar index.

### Expected Behavior

Expected behaviour is the p5.js-website loads all the references from Utils.js and SoundFile of p5.sound.js and show it properly on the p5.sound.js reference page.

### Steps to reproduce

Check out dave's comment.

Yes

@davepagurek @limzykenneth @ksen0 

please take a look whenever you are free.

Thanks a lot for your time.

kunstewi:
fixes #1226 and [#72 (p5.sound.js)](https://github.com/processing/p5.sound.js/issues/72)

@ksen0 @limzykenneth please take a look whenever you are free.

Thanks a lot for your time.

### Old Script

```js
      // Fix p5.sound classes
      for (const key in soundData.classes) {
        const newName = `p5.${  soundData.classes[key].name}`;
        const updated = {
          ...soundData.classes[key],
          name: newName,
        };
        soundData.classes[newName] = updated;
        delete soundData.classes[key];
      }
      for (const item of soundData.classitems) {
        item.class = `p5.${  item.class}`;
      }
```

### Problems

1. In the p5.sound.js repository, global library functions like loadSound(), getAudioContext(), and userStartAudio() are documented with the tag @for p5.sound. YUIDoc reads this information and assigns the functions to the p5.sound class. The old script added p5. to everything without thinking, which meant it took those methods and assigned them to a class called p5.p5.sound that does not exist.

2. The old script mistakenly added p5. to the front of every class and item name, regardless of the situation. This works for a class named SoundFile, which becomes p5.SoundFile. However, if a developer specifically wrote @class p5.SoundFile in the JSDoc comments, the old script would incorrectly change it to p5.p5.SoundFile.

3. The old script used a for...in loop to iterate over the keys inside soundData.classes. In JavaScript, adding and deleting keys in an object while looping through it with a for...in loop is very risky. The JavaScript compiler may become confused. It could end up iterating over the newly added keys in an infinite loop, skipping existing keys, or generating errors.

### Updated Script
```js
      // Fix p5.sound classes and map global methods to the 'p5' class
      const classKeys = Object.keys(soundData.classes);
      for (const key of classKeys) {
        let newName = soundData.classes[key].name;
        if (newName === 'p5.sound') {
          newName = 'p5';
        } else if (!newName.startsWith('p5.')) {
          newName = `p5.${newName}`;
        }

        if (newName !== soundData.classes[key].name) {
          const updated = {
            ...soundData.classes[key],
            name: newName,
          };
          soundData.classes[newName] = updated;
          delete soundData.classes[key];
        }
      }
      for (const item of soundData.classitems) {
        if (item.class === 'p5.sound') {
          item.class = 'p5';
        } else if (!item.class.startsWith('p5.')) {
          item.class = `p5.${item.class}`;
        }
      }
```

this updated script fixes these three critical bugs:

1. Mapping Global Functions to the p5 Sketch Object (The p5.sound to p5 change):

```js
if (item.class === 'p5.sound') {
  item.class = 'p5';
}
```
The new script specifically hunts for anything assigned to the p5.sound class and explicitly reassigns it to the p5 class. This correctly integrates them as global methods in the documentation (alongside things like createCanvas() and ellipse()) instead of breaking them.

2. Preventing "Double Prefixing" (startsWith('p5.'))

```js
} else if (!item.class.startsWith('p5.')) {
  item.class = `p5.${item.class}`;
}
```
The new script safely double-checks if the string already starts with "p5.". If it does, it skips it; if it doesn't, it adds it. This prevents double-prefixing.

3. Safe Object Iteration (Object.keys())

```js
const classKeys = Object.keys(soundData.classes);
for (const key of classKeys) {
```
The new script uses Object.keys() to grab a strict, immutable array of the original class names before the loop starts. It then safely loops over that array, ensuring that any modifications made to soundData.classes down below have no effect on the loop's logic.

#### Previously no result for userStartAudio or any methods in Utils.js and also no loadSound method from SoundFile
Screenshot 2026-03-10 at 5 05 10 PM

#### After the change all methods are present in the reference docs
Screenshot 2026-03-10 at 5 08 08 PM


src/scripts/parsers/reference.ts
```diff
@@ -71,18 +71,31 @@ export const parseLibraryReference =
...
-          ...soundData.classes[key],
-          name: newName,
-        };
-        soundData.classes[newName] = updated;
-        delete soundData.classes[key];
+      // Fix p5.sound classes and map global methods to the 'p5' class
+      const classKeys = Object.keys(soundData.classes);
+      for (const key of classKeys) {
+        let newName = soundData.classes[key].name;
+        if (newName === 'p5.sound') {
```
ksen0: Is there some way to factor out the hardcoded strings here into a const? Both `'p5.sound'` and `'p5'` prefixes are repeatedly used and there's some delicate logic in the whole setup (not this PR specifically, but overall)

ksen0 (changes requested):
Thanks for the fix! I checked that it works locally, just minor note on `p5.sound` string -> const for readability

> The old script used a for...in loop to iterate over the keys inside soundData.classes. In JavaScript, adding and deleting keys in an object while looping through it with a for...in loop is very risky. The JavaScript compiler may become confused. It could end up iterating over the newly added keys in an infinite loop, skipping existing keys, or generating errors.

Just to call this out, thanks for both the catch and the fix.

src/scripts/parsers/reference.ts
kunstewi: Thanks a lot @ksen0 

understood.

done

kunstewi:
added the requested changes.

ksen0 (approved):

Issue #1233, opened 2026-03-12
build: enforce node version requirement clearly when running npm install


nbogie:
### Increasing Access

When developers run the website server or scripts on an unsupported (e.g. too-old) version of node.js they can encounter cryptic, seemingly unrelated problems in the code-base which can waste a lot of their time and be an off-putting developer experience especially for newcomers.

In contrast a system that clearly communicates to them with an error state that their platform is incorrect, allows them to move forward without confusion, saving time and ensuring a more reliable (and welcoming) developer experience.

### Most appropriate sub-area of p5.js?

Other (specify if possible)

### Feature request details

Running `npm install` (or `npm ci`) should fail with a clear descriptive error if an unsupported version of node.js is being used.  The error should state that the version of node.js is unsupported by the software and state what version(s) of node.js ARE acceptable.

### An error, not a warning:

A warning is insufficient as it will very easily be missed in the noisy logs.  Instead, the process should stop with the error being the last thing reported, and the process exit code should be non-zero to allow use in automated setups.

### Support the widest user-base

The solution (i.e. the detection and error mechanism) should work for the widest possible range of developer setups.  For example, whilst specifying `devEngines` in `package.json` will _eventually_ be a solution, at the moment there will still be developers who are using old systems which will ignore this new specification.

Similarly, we cannot assume that all developers are using volta, nvm, or fnm.

### "Why the install command, why not point X?"

The website system consists of many entry points.  Rather than try to cater for all of them, it is simpler and more reliable to address just once the central point that all developer workflows should be including periodically, `npm install`.

### Other package managers:

While ideally the same solution would work with yarn and pnpm, this is not required and we should not add extra files to support those tools.

### Support for CI

It'll be necessary to check that these checks don't _break_ current CI (testing _and_ deployment).

I suggest CI enforcement of the node version requirement can be implemented later.

If, in the solution for the non-CI case, the declaration of supported node version(s) can be read easily by CI, it will allow all systems to eventually run from a single source of truth.

### Consistency with p5.js main library

If possible, it would provide a simpler developer experience if node version requirements for p5.js-website and p5.js development can be the same.

### Documentation

The [prerequisites section in contributor guidelines](https://github.com/processing/p5.js/blob/dev-2.0/contributor_docs/contributor_guidelines.md#prerequisites) should be updated to state the acceptable node.js version range.  

Potentially, a pointer could also be added there so that the developer knows where to look for the up-to-date accepted version range.

[...]

PR #1234, opened 2026-03-12
Enforce node version >= 22 for npm install

nbogie:
Addresses #1233 

* Specify node version >= 22 in [engines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) in `package.json`
* Add .npmrc to have npm error rather than warn when this is not met, using [`engine-strict=true`](https://docs.npmjs.com/cli/v11/using-npm/config#engine-strict) 
* Mentioned this version requirement in `docs/technical_overview.md`.

As a result, npm install with a sub-22 version of node.js will give the following format of error:

```
npm error code EBADENGINE
npm error engine Unsupported engine
npm error engine Not compatible with your version of node/npm: p5.js-website@0.0.1
npm error notsup Not compatible with your version of node/npm: p5.js-website@0.0.1
npm error notsup Required: {"node":">=22.0.0"}
npm error notsup Actual:   {"npm":"10.6.2","node":"v18.12.0"}
```

### why not devEngines?

Whilst [devEngines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) is a good solution in _future_, it will be ignored by older platforms.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,57 +6,39 @@ on:
 permissions:
   contents: read
 
+.common-setup: &common-setup
+  - name: Checkout your repository using git
+    uses: actions/checkout@v4
+  - name: Setup Node.js
+    uses: actions/setup-node@v4
+    with:
+      node-version: '22'
+  - name: Install dependencies
+    run: npm ci  
+
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout your repository using git
-        uses: actions/checkout@v4
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '22'
-      - name: Install dependencies
-        run: npm ci
+      *common-setup
       - name: Test
         run: npm run test
   lint:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout your repository using git
-        uses: actions/checkout@v4
-      - name: Setup Node.js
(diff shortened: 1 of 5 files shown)


nbogie:
Website CI was using node v20 by default (it's not specified).  I've updated it to explicitly use v22 for all steps in test.yml so far, which lets this PR pass. 

Performance note: It'll be worth comparing the CI elapsed run-times for this PR against others to check it is no slower than when it was using  the default from `runs-on: ubuntu-latest`.  Investigate caching options there if it's significantly slower.

update: from the log, this is a good sign:
```
Run actions/setup-node@v4
Found in cache @ /opt/hostedtoolcache/node/22.22.0/x64
```

ksen0:
Thanks, merging! (Is there a similar PR like that should target main?)

From your original issue: "a pointer to guide(s) on how to update version (e.g. for nvm)," that would be helpful, though the error output is relatively clear (below is the tail end of output when attempting to install with old version)

```sh
npm error code EBADENGINE
npm error engine Unsupported engine
npm error engine Not compatible with your version of node/npm: p5.js-website@0.0.1
npm error notsup Not compatible with your version of node/npm: p5.js-website@0.0.1
npm error notsup Required: {"node":">=22.0.0"}
npm error notsup Actual:   {"npm":"11.2.0","node":"v20.17.0"}
npm error A complete log of this run can be found in: [filepath]
```

PR #1235, opened 2026-03-13
added descriptive CDN instruction to download page (#1229)

### LINKED ISSUE #1229: p5 2.x Download Page: Descriptive instruction on p5.js CDN
I'm hoping to set up a local environment that would let me use the latest version of p5.js 2.x without having to look it up manually. 

After learning more about semantic versioning, I think the [download page](https://beta.p5js.org/download/) might benefit from a slightly more descriptive CDN instruction. 

The current description:

>### Use p5.js from CDN
>p5.js is a client-side JS platform that empowers artists, designers, students, and anyone to learn to code and express themselves creatively on the web. It is based on the core principles of Processing.
>`https://cdn.jsdelivr.net/npm/p5@2.2.2/lib/p5.min.js`

Suggested description:

### Use p5.js from CDN
All versions of p5.js are hosted on content delivery network (CDN), a geographically distributed group of servers. There are a couple different ways to access the JavaScript library from your code editor.

#### Current Version of p5.js (recommended)
Use the following URL if you'd like to access the current version of p5.js. 
`https://cdn.jsdelivr.net/npm/p5@2.2.2/lib/p5.min.js`

#### Latest Version of p5.js
Use the following URL if you'd like to ensure that you're always accessing the latest version of p5.js 2.x.
`https://cdn.jsdelivr.net/npm/p5@2/lib/p5.min.js`

#### Other Versions of p5.js
If you'd like to access an earlier version of p5.js, you can look up the corresponding URL on [p5.js' jsdelivr webpage](https://www.jsdelivr.com/package/npm/p5?tab=files).

geeta102:
Resolves #1229

### DIFF
--- a/src/content/text-detail/en/download.mdx
+++ b/src/content/text-detail/en/download.mdx
@@ -8,7 +8,8 @@ import {
   fullDownloadUrl,
   libraryDownloadUrl,
   minifiedLibraryDownloadUrl,
-  p5LatestUrl
+  p5LatestUrl,
+  p5MajorVersion
 } from '@src/globals/globals'
 
 Welcome! This page contains the links to start using p5.js in the way that suits you best. Open the p5.js Editor in your web browser, or download the library to your own computer. We’ve tried to order the links to reflect what a beginner might want first, then what a more experienced programmer may be looking for.
--- a/src/content/text-detail/en/download.mdx
+++ b/src/content/text-detail/en/download.mdx
@@ -43,7 +43,7 @@ Use the following URL if you'd like to access the current version of p5.js.
 
 ### Latest Version of p5.js
 
-Use the following URL if you'd like to ensure that you're always accessing the latest version of p5.js 2.x.
+Use the following URL if you'd like to ensure that you're always accessing the latest version of p5.js {p5MajorVersion}.x.
 
 <CodeContainerWithCopy>{p5LatestUrl}</CodeContainerWithCopy>
 
--- a/src/content/text-detail/en/download.mdx
+++ b/src/content/text-detail/en/download.mdx
@@ -7,7 +7,7 @@ import {
   cdnLibraryUrl,
   fullDownloadUrl,
   libraryDownloadUrl,
-  minifiedLibraryDownloadUrl
+  minifiedLibraryDownloadUrl,
   p5LatestUrl
 } from '@src/globals/globals'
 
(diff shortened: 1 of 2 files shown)


src/globals/globals.ts
```diff
@@ -26,3 +26,5 @@ export const cdnSoundUrl =
   (!!import.meta.env?.PUBLIC_P5_LIBRARY_PATH || p5Version.startsWith('v2'))
     ? `https://cdn.jsdelivr.net/npm/p5.sound@${p5SoundVersion}` as const
     : `https://cdn.jsdelivr.net/npm/p5@${p5Version}/lib/addons/p5.sound.min.js` as const
+export const p5LatestUrl =
```
ksen0, line 31: Please update this to be not hardcoding `2.x` but rather using `p5Version` from above to just get the major version.

When p5Version is 2.x.x, then this should be p5@2. However if `p5Version` is 1.x.x, `p5LatestUrl` should be using that major version and be p5@1

src/content/text-detail/en/download.mdx
```diff
@@ -31,10 +32,21 @@ These are downloads or links to the p5.js library file. No additional contents a
...
 
 <CodeContainerWithCopy>{cdnLibraryUrl}</CodeContainerWithCopy>
 
-## Older versions
-Looking for older versions?
-Head over to GitHub for [earlier releases](https://github.com/processing/p5.js/releases) of p5.js.
\ No newline at end of file
+### Latest Version of p5.js
+
+Use the following URL if you'd like to ensure that you're always accessing the latest version of p5.js 2.x.
```
ksen0: Same here, although the original issue gave 2.x as an example, this works for any version of p5. So when p5Version is 1.x this should be "Use the following URL if you'd like to ensure that you're always accessing the latest version of p5.js 1.x." This should also be calculated from p5Version and not hardcoded

ksen0 (changes requested):
Thanks @geeta102 ! A couple of comments, to make this PR a bit more general. Right now it targets main brach, but on main this logic has to be about version 1.x . 

Once this is in, I will also apply it in the `2.0` branch so it is on both versions of the site!

src/globals/globals.ts
geeta102: done, updates pasteUrl and used p5Version.split('.')[0] instead of hardcoded 2.x

src/content/text-detail/en/download.mdx
geeta102: Done,  Updated the text to use {p5MajorVersion}.x instead of hardcoded 2.x

doradocodes:
@geeta102, @Nwakaego-Ego and I have reviewed your change with @ksen0's feedback and it looks good! As a final step, please update your branch with the latest version of `main`.

ksen0 (approved):

PR #1237, opened 2026-03-15
Prepare for WebGPU docs

davepagurek:
I've started writing some experimental compute shader reference docs in https://github.com/processing/p5.js/pull/8531. WebGPU docs are a little different from regular ones because you need to include an extra addon script. They're a little like p5.sound in that sense, but unlike p5.sound, which has its own reference page to itself, that will *probably* not be the case for WebGPU; instead, they will likely be mixed in with other categories, like p5.strands.

So this PR:
- Updates the `custom:dev` script to also copy over the webgpu addon from the branch it builds from
- Updates the reference builder to copy over the `webgpu?: boolean` property of reference data, which will indicate that the item will need the WebGPU script in its examples
- Updates the UI components to pass in the WebGPU addon script if necessary
- Updates the reference builder to copy over the `webgpuOnly?: boolean` property of reference data, which will indicate that the item is only available in the WebGPU addon. (e.g. some examples of `model()` may use webgpu, but the method itself doesn't need the addon; `buildComputeShader` *needs* the webgpu addon.)
- Adds a warning to WebGPU reference items saying that you need an extra addon script to use them.

Here's a preview:
image

The double-warning of experimental + webgpu isn't super great UI, let me know if anyone has ideas on how to make that nicer!

### DIFF
--- a/src/components/CodeEmbed/index.jsx
+++ b/src/components/CodeEmbed/index.jsx
@@ -39,7 +39,7 @@ export const CodeEmbed = (props) => {
   );
 
   let { previewWidth, previewHeight } = props;
-  const canvasMatch = /createCanvas\(\s*(\d+),\s*(\d+)\s*(?:,\s*(?:\w+\.)?(?:P2D|WEBGL)\s*)?\)/m.exec(initialCode);
+  const canvasMatch = /createCanvas\(\s*(\d+),\s*(\d+)\s*(?:,\s*(?:\w+\.)?(?:P2D|WEBGL|WEBGPU)\s*)?\)/m.exec(initialCode);
   if (canvasMatch) {
     previewWidth = previewWidth || parseFloat(canvasMatch[1]);
     previewHeight = previewHeight || parseFloat(canvasMatch[2]);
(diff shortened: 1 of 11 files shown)


nbogie:
I think the double warning is good. That's two separate things to have attention drawn to.

davepagurek:
Do you think the warning icon is appropriate here? It's probably fine, although I wonder if it might make sense to try to make it look less like the deprecation warning, which is what it's currently borrowed from. Something maybe more inviting (different colour coding?) could be good, although this is probably a fine starting point too.

nbogie:
While a jigsaw-piece icon or building-blocks icon _could_ signify an add-on (if there's not an icon in use for that yet, the warning icon already best says to a lot of readers: "Hey!  Watch out, this code probably isn't going to work for you unless X", and is easily the best for getting people's attention so that they avoid frustration and wasted time.  
For it to be anything otherwise risks that people just think it's some cute extra info they don't need to read.  
There aren't so many instances of the deprecation warning throughout the docs - i don't think people are desensitised to it.  I just read it as a warning.
I might argue the order of those warnings should be switched  to lessen the chance the "addon-needed" warning is overlooked, because I think of the two, missing that one is likely to waste more time. Very minor.

davepagurek:
Makes sense! swapped the order.

PR #1238, opened 2026-03-16
Add p5.strands translations for shader examples and rename original GLSL examples

MissTipo:
## Summary

This PR adds p5.strands translations for three existing shader examples and keeps the original GLSL versions alongside them for comparison.

## Changes

- renamed the original examples:
  - `Filter Shader`
  - `Adjusting Positions With A Shader`
  - `Shader As A Texture`
- added p5.strands translations for each of the above examples
- updated and remixed the example descriptions to match the renamed originals and the new p5.strands versions

## Notes

The goal was to preserve the original GLSL examples while adding equivalent p5.strands versions that are easier to compare side by side.

## Testing

- ran the examples locally in the web editor/site preview
- verified that the new p5.strands examples render correctly
- verified that the original examples remain available under their renamed versions


src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
```diff
@@ -0,0 +1,81 @@
...
+  // baseMaterialShader is used because it is the only base shader whose
+  // getPixelInputs hook exposes inputs.texCoord in JS callback form.
+  // baseColorShader only exposes getFinalColor, which does not include texCoord.
+  theShader = baseMaterialShader().modify(shaderAsTextureCallback);
+
+  describe('Sphere broken up into a square grid with a gradient in each grid.');
+}
+
+function shaderAsTextureCallback() {
+  const time = uniformFloat(() => millis() / 1000.0);
```
davepagurek: As in the earlier example, this doesn't actually need to be a uniform any more, inside of `pixelInputs` you should be able to write `const time = millis() / 1000)`.

src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
```diff
@@ -0,0 +1,81 @@
...
+  const size = uniformVec2(() => [width, height]);
```
davepagurek: I believe we can also just use `width` and `height` directly now without uniforms too!

src/content/examples/en/11_3D/07_Filter_Shader_p5strands/code.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function displaceColorsCallback() {
+
+  // Note: arithmetic on coord operates componentwise (vec2), as in GLSL.
+  function zoom(coord, amount) {
+    return (coord - 0.5) / amount + 0.5;
+  }
+
+  getColor((inputs, canvasContent) => {
```
davepagurek: We've recently introduced a flatter p5.strands API that uses `filterColor.begin()` and `filterColor.end()` instead of `getColor` with a callback, since people have found the nested functions confusing. See some examples in these docs: https://beta.p5js.org/reference/p5/buildfiltershader/

Mind switching to that syntax in these examples? for most other `get*` hooks, the `get` part is just removed, e.g. `getWorldInputs` becomes just `worldInputs`.

src/content/examples/en/11_3D/08_Adjusting_Positions_With_A_Shader_p5strands/code.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+    }
+    endShape();
+  });
+
+  describe('A red-to-blue ribbon that waves over time');
+}
+
+function wiggleCallback() {
+  const time = uniformFloat(() => millis());
```
davepagurek: I believe this doesn't actually need to be a uniform any more, you can just use `millis()` directly inside of a strands shader 🙂

src/content/examples/en/11_3D/08_Adjusting_Positions_With_A_Shader_p5strands/code.js
```diff
@@ -0,0 +1,46 @@
...
+
+  getObjectInputs((inputs) => {
```
davepagurek: Similar to before, can we use the flat API here?

src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
```diff
@@ -0,0 +1,81 @@
...
+  // Y is flipped to match gl_FragCoord's bottom-up axis.
```
davepagurek: Partially a comment for @ksen0: do you think the main target audience of these examples is people looking to understand strands from a background of GLSL, or people just learning about strands who don't already know GLSL? For the latter, explaining how things translate to `gl_FragCoord` is introducing a point of reference that they aren't familiar with. If that's the goal, it might be better to just omit references to how `gl_FragCoord` worked in the old version of this?

src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
```diff
@@ -0,0 +1,81 @@
...
+  const mouse = uniformVec2(() => [mouseX, height - mouseY]);
```
davepagurek: Similarly, I think `mouseX` and `mouseY` should work without a uniform too! `const mouse = [mouseX, mouseY]` in a shader hook block should work.

src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
```diff
@@ -0,0 +1,81 @@
...
+
+  function tile(st, zoom) {
+    return fract(st * zoom);
+  }
+
+  function concentricCircles(st, radius, res, scale) {
+    return floor(distance(st, radius) * res) / scale;
+  }
+
+  getCameraInputs((inputs) => {
```
davepagurek: Can we switch these to the begin/end syntax?

src/content/examples/en/12_Advanced_Canvas_Rendering/03_Shader_As_A_Texture_p5strands/code.js
davepagurek: In p5.strands we also don't need the `.0` on numbers, `1000` will be fine.

davepagurek (commented):
Nice work converting these! This all works, so I just left some comments more about code style and using the most recent p5.strands features to simplify the code.

MissTipo:
> Nice work converting these! This all works, so I just left some comments more about code style and using the most recent p5.strands features to simplify the code.

Thanks Dave!  Making the changes. Will push updates soon.

src/content/examples/en/11_3D/08_Adjusting_Positions_With_A_Shader_p5strands/description.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+  - description: Remixed by
+    attribution:
+      - name: Dorine Tipo
+        URL: https://dorinetipo.vercel.app/
+    code:
+      - label: 2026 p5.strands translation
+        URL: https://github.com/processing/p5.js-website/tree/main/src/content/examples/en/11_3D/08_Adjusting_Positions_With_A_Shader_p5strands
+
+  - collectivelyAttributedSince: 2024
```
ksen0: Hi Dorine! Please update this to 2026. So basically it's collectively attributed after it's published from you.

src/content/examples/en/12_Advanced_Canvas_Rendering/02_Shader_As_A_Texture/description.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 featuredImage: "../../../images/featured/12_Advanced_Canvas_Rendering-02_Shader_As_A_Texture-thumbnail.png"
 featuredImageAlt: Two spheres broken up into a square grid with a gradient in each grid.
-title: Shader as a Texture
+title: Shader as a Texture(GLSL)
```
ksen0: Minor: please add space before (

ksen0:
These look wonderful thank you @MissTipo !

Some minor notes added.

I was also inspecting how it looks on the website, and I propose that we add the 3 strands sketches to featured; and remove 3 of the previous sketches except "orbitControl." This would bring more visibility to the strands examples. Other ideas on how to make this more navigable is welcome @davepagurek (also @nbogie maybe you have thoughts?)

image

image

For implement featured solution: add/remove `featured: true` from the description mdx of the affected sketches.

Finally, all past sketches do have translations. Minor/can be done in a separate PR but everywhere were the description text is the same, the existing translations can be duplicated for the strands sketch, I think.

ksen0 (changes requested):
Minor

src/content/examples/en/11_3D/08_Adjusting_Positions_With_A_Shader_p5strands/description.mdx
ksen0: Also in the other places where this appears

MissTipo: Hi Kit! Quick heads up: changing collectivelyAttributedSince to 2026 needs a couple updates beyond the description.mdx files, otherwise Astro fails schema validation:
1. `src/content/examples/config.ts` — the schema needs to be updated to accept the new year value
2. `src/content/ui/*.yaml` files — a new attribution string needs to be added for the 2026 year

I've made both changes. Should I include them in this PR, or open a separate one?

ksen0: I'm okay with it being in this PR, I think it should be <20 lines? If more then maybe separate

MissTipo:
> Finally, all past sketches do have translations. Minor/can be done in a separate PR but everywhere were the description text is the same, the existing translations can be duplicated for the strands sketch, I think.

Happy to handle the translation copying too! A quick clarification on titles, should I append (p5.strands) to the title in each new strands translated file, and correspondingly append (GLSL) to the title in each existing translated file as with the English description or should everything stay as is?

ksen0:
Thanks @MissTipo ! Yes please update the titles to include the (GLSL) and (p5.strands). Thank you!

cc @Divyansh013 for visibility - I think for GLSL and p5.strands those are technical terms and can stay in English but please let me know if there are any other ideas you have?

davepagurek (commented):
The code updates look great, nice work @MissTipo!

ksen0 (approved):

PR #1241, opened 2026-03-16
beta_deploy: have astro build action use node 22

nbogie:
Addresses: #1233 

Changes beta_deploy.yml build-deploy job to explicitly tell the astro build action which version of node.js to use.  

This is necessary because that astro build action will do an `npm install` which will fail if it's not at least on node 22.

The action's [documentation is here](https://github.com/withastro/action/tree/v2) (for v2 of the action).  (I note v5 of the action already defaults to run on node.js v24, but it's better we state it explicitly so maintainers are aware of what's being used and can easily control it).

### DIFF
--- a/.github/workflows/beta_deploy.yml
+++ b/.github/workflows/beta_deploy.yml
@@ -26,6 +26,8 @@ jobs:
         env:
           NODE_OPTIONS: "--max_old_space_size=4096"
         uses: withastro/action@v2
+        with:
+          node-version: 22
         
       - name: Deploy to Netlify
         uses: nwtgck/actions-netlify@v3.0


nbogie:
Testing update: 
1. on my own github, tested a simplification of this beta_deploy.yml which skips the final netlify deploy.  It correctly allows the astro build (v2) to run successfully on node 22.
2. tried _without_ the node-version stipulation, and it breaks as expected (it is defaulting to node 20 and noting the mismatch against what package.json and .npmrc dictate)
3. Successfully tested the full deploy to netlify, too. (well, it built and deployed ok then failed because it didn't have perms to update my repo.  unrelated.)

doradocodes:
@nbogie Wondering if it's useful to also specify the Node version in the package.json?

nbogie:
Hi @doradocodes ! 

> specify the Node version in the package.json?

Yes, that's done in the recently merged #1234 

There's some more detail in the main issue: #1233 

Thank you for recommending it!

Issue #1256, opened 2026-03-18
2.0: Add a simple custom 404 to help with broken reference page links in github


nbogie:
### Increasing Access

There are many beta reference page links in github issues that have now-incorrect casing (also in discord, elsewhere).

Example of problem:
This issue https://github.com/processing/p5.js/issues/8623
has links to...
https://beta.p5js.org/reference/p5.font/texttocontours/
https://beta.p5js.org/reference/p5.font/texttomodel/
These now needs to be camel case, like the main site:
https://beta.p5js.org/reference/p5.font/textToContours/
https://beta.p5js.org/reference/p5.font/textToModel/

When the reader clicks such an incorrect URL, they are currently presented with a raw 404 message from their browser, with no assistance.  The experience is unhelpful - most users will not know what has gone wrong.

<figure>
<img width="593" height="345" alt="Image" src="https://github.com/user-attachments/assets/806befaf-1fff-4c6a-bd57-16b1977f105a" alt="an example message from browser reporting that the page was not found.  It gives no other assistance." />
  <figcaption><br/>An example unhelpful message from chrome browser</figcaption>
</figure>

### Most appropriate sub-area of p5.js?

Other (specify if possible)

### proposal

Add a custom simple 404 page which presents a link to allow the user to quickly and automatically search the website for the name of the page that wasn't found.

* It should search using the website
* In most cases it might be enough to search only for the last part of the slash-separated path
  * `/reference/p5.font/texttocontours/` would search for `texttocontours`

#### Example:

Visiting the bad URL: https://beta.p5js.org/reference/p5.font/texttocontours/
should present a 404 page that includes a link to: https://beta.p5js.org/search/?term=texttocontours

### other work

Note: this issue doesn't replace the planned 404 pages redesign: #1049.

The intention is to put in place something simple _quickly_ which will minimise disturbance to users.  

Eventually there will be other solutions to this problem of case-mismatching links (e.g. perhaps redirection on server).

### beta site only

Note: this issue applies only to the beta site.

nbogie:
I'd like to work on this.

I have a PR ready, assuming cloudflare can look for and use `src/pages/404.astro` or similar.

nbogie:
I note i should also strip any hash part from the url path before using it in search

`https://beta.p5js.org/reference/p5.font/texttocontours/#Examples`
should use
`texttocontours`


Done

ksen0:
Associated PR is closed

PR #1257, opened 2026-03-18
Add rough custom 404 page in astro with quick search

nbogie:
Addresses #1256 

### changes

1. adds custom 404 page with populated search link
2. configures cloudflare to look for it

#### 1. custom 404 page with populated search link

Adds a rough custom 404 page in astro (`/src/pages/404.astro`) which presents user with a quick search option to search the website (using the website) for pages matching the last part of their path.

Example:
When the user visits the broken URL: https://beta.p5js.org/reference/p5.font/texttocontours/
the custom 404 page includes a search link to: https://beta.p5js.org/search/?term=texttocontours

image

#### 2. Cloudflare config

per the astro docs for cloudflare deployment
https://docs.astro.build/en/guides/deploy/cloudflare/#404-behavior

this PR changes `wrangler.jsonc` to  set `not_found_handling` as follows:

```json
"assets": {
    "directory": "dist",
    "not_found_handling": "404-page"
  },
  ```
 
This will cause it to look for a 404 page in `.dist/404.html`.  I have confirmed locally astro build is putting our custom 404 page at that file path.

(Note to anyone testing with other their own astro stuff: the popular "starlight" theme will overwrite this page by default with its own 404 page.  (this can be disabled.))

### DIFF
--- a/.github/workflows/beta_deploy.yml
+++ b/.github/workflows/beta_deploy.yml
@@ -20,6 +20,10 @@ jobs:
     steps:
       - name: Checkout your repository using git
         uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
       - name: Install dependencies
         run: npm ci
       - name: Build website
--- a/.github/workflows/beta_deploy.yml
+++ b/.github/workflows/beta_deploy.yml
@@ -1,4 +1,4 @@
-name: Deploy beta to Netlify
+name: Deploy beta to Cloudflare
 
 on:
   # Trigger the workflow every time you push to the `main` branch
@@ -20,6 +20,8 @@ jobs:
     steps:
       - name: Checkout your repository using git
         uses: actions/checkout@v4
+      - name: Install dependencies
+        run: npm ci
       - name: Build website
         run: npm run build
       - name: Deploy to Cloudflare Workers
--- a/src/pages/404.astro
+++ b/src/pages/404.astro
@@ -1,9 +1,8 @@
 ---
 import Head from "@components/Head/index.astro";
 import BaseLayout from "../layouts/BaseLayout.astro";
-import { getCurrentLocale, getUiTranslator } from "../i18n/utils.ts";
+import { getCurrentLocale } from "../i18n/utils.ts";
 const currentLocale = getCurrentLocale(Astro.url.pathname);
-// const t = await getUiTranslator(currentLocale);
 ---
 
 <Head title="404: Page Not Found" locale={currentLocale} />
--- a/src/pages/404.astro
+++ b/src/pages/404.astro
@@ -1,0 +1,52 @@
+---
+import Layout from "../layouts/BaseLayout.astro";
+
+/**
+ * Returns the last (non-empty) segment from the slash-separated path, or ""
+ * 
+ * @param path string to extract the last part of
+ * @returns the last non-empty segment or an empty string
+ * 
+ * @example 
+ * an input of `/reference/p5/basematerialshader/` will return `basematerialshader`
+ */
+function getLastPathElementOrEmptyString(path: string): string {
+    return path.split("/").filter(Boolean).pop() ?? "";
+}
+
+
+const currentPath = Astro.url.pathname;
+const finalTerm = getLastPathElementOrEmptyString(currentPath);
+---
+
+<Layout title="404: Page Not Found">
+    <main class="error-container">
+        <p>Sorry, we couldn't find your exact page <code>{currentPath}</code>.</p>
+
+        <div class="actions">
+            <a
+                href={`https://beta.p5js.org/search/?term=${finalTerm}`}
+                class="btn secondary"
+            >
+                Search here for <code>{finalTerm}</code>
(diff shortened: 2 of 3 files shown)


ksen0:
@nbogie for the config, I think that goes in here? If I understand correctly, can be addable to this PR. https://github.com/processing/p5.js-website/blob/2.0/wrangler.jsonc

ksen0:
Do you have any ideas on how to get the user's browser from insisting on lowercase? That seems like a pretty big problem even with 404 implemented

nbogie:
> Do you have any ideas on how to get the user's browser from insisting on lowercase? That seems like a pretty big problem even with 404 implemented

I'm hopeful that it seems this correction only happens when the user tries hand-correcting the capitalisation of an all-lower-case url in the url bar, and then presses enter, when their browser "knows better" (or can't differentiate wrt to cache lookup).  It **doesn't** seem to happen if the user **clicks** a link with a corrected capitalisation - this seems to bypass any browser correction, if indeed that's what's happening in the former case.
 
### Steps to repeat: (on chrome)

1. open chrome://history 
1. search for `beta.p5js.org/reference`
1. find a multi-word page name that is (or should be) camel-cased (it will be displayed correctly, in history!)
1. click it
1. observe a 404 due to all-lowercase url path
1. try to fix the url capitalisation manually, and press enter
1. observe chrome adjusts your edit to lowercase so you get the 404 again

As for preventing this relatively minor correcting, in chrome it looks like it is only possible with "clear browsing history". https://stackoverflow.com/questions/50298340/chrome-changes-url-case
The couple of test cases I was using, I think i fixed by removing the specific search suggestion on the url bar (x there).
(There used to be a way to turn off search autocomplete in chrome, but not now, it seems.)

nbogie:
### Testing:

* tested the 404 search locally
* tested the cloudflare 404 config by building and deploying a tiny astro project of my own on cloudflare using github CI wrangler action
  * limitation: not p5js-website repo. But similar config wrt 404.

Update: now using wrangler action on github actions, not wrangler manually from cli

nbogie:
Good to go, i think!

ksen0:
Thanks so much for your work on this ! It did look reasonable and I did deploy, however the routes are doing a different thing than expected. Here is me clicking on "curveDetail"* and getting into weird 404:

https://github.com/user-attachments/assets/63469209-1555-44ec-9d4f-59f86bb33e0d

Reproduced with both /curveVertex/ and /curveVertex in path. But feels like surely it's related to trailing slash handling in dev vs prod

I have now redeployed from `2.0` without this patch.

*"It doesn't seem to happen if the user clicks a link with a corrected capitalization - this seems to bypass any browser correction, if indeed that's what's happening in the former case." Alas, not so in general. In Chrome I consistently get the browser talking to itself loop if I click on something I've clicked on before / which is cached.

nbogie:
That looks like clicking on a _search result_ was taking you to the 404 page?

  *And* (perhaps) the 404 page wasn't presenting any part of your original link in the search link? (Or perhaps when you get taken to 404 it is snapping to a certain url).

nbogie:
So, the JS in the 404 astro page was getting statically compiled and the values locked in from the first render during build.  My bad for not test-deploying the full 404 in my toy astro site.

I have successfully recreated the problem now with a cloudflare deploy of the the website repo and will change it to use JS that definitely runs client-side instead!

nbogie:
Fixed!  You can test it out here: https://p5-js-website-beta.neillzero.workers.dev/reference/p5/basestrokeshader/ 

(The above is a deploy of a branch that's very slightly different from this PR branch, with (only) some necessary removals of custom p5.js domains from the wrangler config)

nbogie:
In my tests I can't recreate the (possible routing) issue shown in Kit's video where the user clicks a **search result** and it goes to the 404 page.

I wonder if that's not actually a routing issue, but this other issue of the browser "correcting" the "wrong" capitalisation of a URL it's been to before.

One easy way to tell is to use a browser that's not been to those URLs before.

ksen0:
> In my tests I can't recreate the (possible routing) issue shown in Kit's video where the user clicks a search result and it goes to the 404 page.

> I wonder if that's not actually a routing issue, but this other issue of the browser "correcting" the "wrong" capitalisation of a URL it's been to before.

Yes I believe that must be it! Was fine in incognito.

ksen0:
image

Something weird with style (actually also on https://p5-js-website-beta.neillzero.workers.dev/reference/p5/buildfiltershader/):

1. Footer is unmoored from the bottom of the page
2. Social arrows incorrect encoding?

nbogie:
I fixed the social arrows by using a Head + BaseLayout , bringing it into line with most of the pages on the site.  But I wasn't able to reproduce the "footer unmoored" issue.

(The Head component sets encoding - that was likely the social arrows issue.)

deployed here: https://p5-js-website-beta.neillzero.workers.dev/reference/p5/buildstrokeshader/

ksen0:
Wonderful, thanks so much! Updating.

The footer floating away from bottom of the window is a normal outcome of page not being long enough to be scrollable:

image

Not introduced in this PR, and anyway 404 design will continue, so will ignore.

Issue #1268, opened 2026-03-22
Improve Translation Tracker: expand content paths and add dry run mode


star11293:
The translation tracker workflow currently only triggers on examples and tutorials content but the script itself supports 6 content types (examples, tutorials, reference, text-detail, events, libraries). The workflow paths should match what the script supports.

Also there's no easy way to test the tracker locally without a GitHub token. A --dry-run flag would let contributors preview what issues would be created without actually creating them.

I have a branch ready with these changes if this sounds good to work on.

ksen0:
Hi @star11293 thanks for the enthusiasm! Please, before future work, review [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/): Issues should be approved and assigned first. PRs made outside of this process may be closed.

I'll let @Divyansh013 decide about this one; however my sense is that the dry run logic differs a bit much from the action. For testing Actions, one could use nektos; or test on one's own fork; maybe the better direction is to document these methods. The translation script already has modes for running locally (see: "Scan All Files" commands in https://github.com/processing/p5.js-website/tree/main/.github/actions/translation-tracker)

Thanks both!

star11293:
@ksen0 Thanks for the feedback :) and noted on the process. Going forward, I'll make sure to wait for an approval before opening PRs

That makes sense about the existing local testing methods. I'll wait to hear from Divyansh on this one.

PR #1269, opened 2026-03-22
Expand translation tracker content paths and add dry-run mode (addresses #1268)

star11293:
Addresses #1268

The workflow was only triggering on examples and tutorials but the script supports 6 content types. So i expanded the paths to cover all of them (reference, text-detail, events, libraries).

Also added a --dry-run flag to index.js so contributors can preview what issues would be created without needing a GitHub token.Then i updated test-local.js with --scan-all and --content-type flags for easier local testing. Lastly, i  cleaned up the README to match.


Divyansh013:
Hi @star11293 , since you have already shown interest on this with your gsoc proposal, I'll be closing this for now. I have mailed you the feedback on your proposed approach.

PR #1278, opened 2026-03-25
Issue #1277

### LINKED ISSUE #1277: Browser scrolls on up and down key inputs on snake example

Examples

Windows

### Web browser and version

Chrome 146.0.7680.81 

### Actual Behavior

On the snake example: https://p5js.org/examples/games-snake/ the browser will scroll when up and down arrow keys are pressed

### Expected Behavior

Once you've started playing snake, these scrolling actions are disorienting. The game should remained fixed. I will comb for other examples that may need a similar treatment.

### Steps to reproduce

_No response_

On it!

bryanhoffman:
…p and down arrow keys are pressed

resolves #1277

Link: https://github.com/processing/p5.js-website/issues/1277

### DIFF
--- a/src/components/CodeEmbed/frame.tsx
+++ b/src/components/CodeEmbed/frame.tsx
@@ -42,6 +42,11 @@ canvas {
 }
 ${code.css || ""}
 </style>
+<script>
+window.addEventListener("keydown", (e) => {
+  if (["ArrowUp", "ArrowDown", " "].includes(e.key)) e.preventDefault();
+}, { passive: false });
+</script>
 <!-- If we need an addon script, load p5 the usual way with no caching to make sure
 the import order doesn't get messed up. -->
 ${((code.scripts?.length ?? 0) > 0 ? [(code.scripts ?? [])] : []).map((src) => `<script type="text/javascript" src="${src}"></script>`).join('\n')}


ksen0:
Thanks for finding this bug and reporting it! (In the future @bryanhoffman it would be really helpful if you waited to be assigned to work on an issue before filing a PR. I know this can require a bit more patience, but it also helps reduce duplicate work or avoidable revisions. Thank you for understanding!)

In this case, I'd like to reconsider the approach of adding a windowlistener. I would like to investigate a bit whether this might be a regression (I don't think so? @davepagurek does this seem like the behavior that's always been there?). If you have ideas on how to fix this in the component that displays the code and the sketch (not the sketch code itself), you are welcome to revise this.

Also cc @doradocodes if you have any ideas related here!

ksen0 (changes requested):
Let's look for a solution that doesn't require changing sketch code itself, if at all possible!

bryanhoffman:
@ksen0 I'm sorry! I thought what I was doing was helpful, but I'm very new to open source contributions. I don't know what happens on your end when I submit a PR. When I submit a PR, what "duplicate work or avoidable revisions" does it create? I appreciate the feedback, I want to be helpful and learn. Sorry for making a mess!

As for fixing this without altering example code, if I run:

```
window.addEventListener("keydown", (e) => {
  if (["ArrowUp", "ArrowDown", " "].includes(e.key)) e.preventDefault();
}, { passive: false });
```

in the console after changing it so it's running in the about:srcdoc context, it fixes the scrolling issue. I've attempted to place the logic in an appropriate place namely the src/components/CodeEmbed/frame.tsx file.

ksen0:
@bryanhoffman no worries at all! n this case "duplicate work" doesn't apply , but we have the guideline against unassigned PRs (in [contributor guidelines](https://p5js.org/contribute/contributor_guidelines/)) because sometimes people make PRs to an issue someone else is already working on, which mean the unassigned PR has to be closed. When I say "avoidable revisions" I mean if code is written before there's an agreement on what actually the fix should be.

In this example, what could have been done differently (what to do in the future): make the issue. Comment on the issue how you would fix it. A maintainer or steward (scroll down to the table [here](https://github.com/processing/p5.js) to see who that would be) would check and reply, assigning you the issue and providing feedback on the approach.

The p5.js project is now over a decade old. The website is newer, but has many complex integrations. Even fixes that seem simple usually aren't and the process suggested in the guidelines is the way it is to try to build in ways to give technical guidance before contributors invest a lot of energy into a solution. I hope that clarifies! And welcome :) !

src/components/CodeEmbed/frame.tsx
```diff
@@ -42,6 +42,11 @@ canvas {
 }
 ${code.css || ""}
 </style>
+<script>
```
ksen0, line 45: It is very important not to make keyboard navigation unusable on the site as a whole: ([WCAG 2.1](https://www.w3.org/WAI/WCAG21/Understanding/keyboard-accessible))

cc website accessibility steward @coseeian  - in this case, the sketch does something on keyboard arrow up/down. I assume the proposed solution, which ignores arrow up and down, would make keyboard navigation not usable. However, if it is possible to only ignore these when there is a sketch "in focus", would that be ok?

I am actually not sure how complex it would be to only do this when a sketch is "in focus" @bryanhoffman, if the current sketch viewer component has such information

bryanhoffman: @ksen0 It should only ignore up and down arrows and the space bar when the sketch is "in focus." If I click the sketch it will start to ignore these keys, but if I then click outside of the sketch, the browser starts to "hear" up, down, and space again. I verified this locally. I think that should be suitable for accessibility concerns.

coseeian:
I’d like to raise a concern regarding the global application of this change.

Intercepting these keys by default may inadvertently break examples that rely on native browser behavior. For instance, in the [DOM Form Elements example](https://p5js.org/examples/input-elements-dom-form-elements/), the `createSelect()` dropdown becomes nearly unusable because the arrow keys are blocked, preventing users from navigating options. This creates a significant accessibility barrier for keyboard-reliant users.

Perhaps we should narrow the scope of this intervention to specific cases that require it (e.g., the Snake Game) while leaving the default iframe behavior untouched.

bryanhoffman:
@coseeian Here's my first PR where I put the logic into the example code itself: https://github.com/bryanhoffman/p5.js-website/commit/ffb29aefd

Do you think this approach is preferrable?

Alternatively, I'm starting to think that this might not be a website issue, but actually something needing ported to the p5.js library when keyIsPressed() or keyPressed() is used. I'm thinking the logic would go approximately here: https://github.com/processing/p5.js/blob/9206f8d542b248d767ecdc652a26de988926952c/src/events/keyboard.js#L443

I'm happy to hear thoughts on this. Thanks to everyone for the help!

bryanhoffman:
Sorry for back to back comments. I want to expound a little more on these couple of options. The more nuclear option is adding something like:

```
// Block the default browser scrolling
const reservedKeys = [32, 37, 38, 39, 40]; // Space, Left, Up, Right, Down
if (reservedKeys.includes(e.keyCode)) {
  e.preventDefault();
}
```

to the p5.prototype._onkeydown function on line 443. 

One more alternative is the creation of a disableScrollKeys() function within the p5.js library. So, this would still rely on developers adding this function to their examples and would result in the example code changing but it seems to embody the friendly philosophy of the p5.js project.

coseeian:
I understand that we should keep the sketch code clean for learners. However, a global kill switch for arrow keys is a no-go for accessibility as well.

What if we added a prop to the CodeEmbed component to toggle this behavior? That way, we only intercept keys for sketches that truly require it (like games) without affecting other examples.

bryanhoffman:
Thanks for the suggestions. Following the discussion, I’ve been exploring an engine-level solution by adding a suppressScrolling() function directly to the p5.js library.

I’ve successfully implemented this locally. By adding the function to the prototype, we can allow sketches to opt-in to scroll prevention.

Implementation Detail:

```
/**
 * The suppressScrolling function
 * Prevents the browser window from scrolling when the user interacts
 * with the canvas using common gaming keys (Arrows, Space, etc.).
 **/
p5.prototype.suppressScrolling = function() {
  const keysToBlock = [
    'ArrowUp',
    'ArrowDown',
    'ArrowLeft',
    'ArrowRight',
    ' ', // Spacebar
    'PageUp',
    'PageDown',
    'End',
    'Home'
  ];

  window.addEventListener('keydown', (e) => {
    if (keysToBlock.includes(e.key) && document.activeElement === this.canvas) {
      // We only prevent scroll if the focus is on the canvas or the sketch
      // This ensures the rest of the page remains accessible
      e.preventDefault();
    }
  }, { passive: false });
};
```

Usage in Example:

```
function setup() {
  createCanvas(400, 400);

  // Draw rectangles from their center
  // This makes it easier to check whether the ball is above or below the
  // center of a paddle
  rectMode(CENTER);
  fill(255);
  noStroke();
  textSize(40);
  textAlign(CENTER);

  // Start paused
  noLoop();
  describe(
    'Two narrow white rectangles and a white square representing the paddles and ball in a game of ping pong. The player scores are displayed in the upper corners, and initially text reads "Click to start"'
  );

  suppressScrolling();
}
```

I’ve verified the fix locally (see attached video). I’m happy to draft the reference documentation for this new method as well. This feels like a clean and reusable way to fix the issue.


https://github.com/user-attachments/assets/6e608c7b-1e91-43e0-9fd9-b2665c226178



I've learned a lot already. Thanks again everyone!

calebfoss:
Hey @bryanhoffman, thank you for your work on this! I have not yet had a chance to run this locally but wanted to bring up a couple thoughts:

- Good thinking on checking that the canvas is focused before bypassing those key inputs. Have you checked that this works in instance mode? If you're unfamiliar, [here is an example](https://p5js.org/examples/advanced-canvas-rendering-multiple-canvases/) that demonstrates it. It looks like the `canvas` property on the p5 instance works the same in both global and instance mode, but it would be worth testing your function in both.
- While this is an issue ultimately worth addressing directly, an option that would be available in the meantime if more discussion/research is needed, using what is already in the API, would be editing the relevant examples to use [fullscreen()](https://p5js.org/reference/p5/fullscreen/). The bonus there would be adding a demonstration of a function that would likely be popular for creators making games.

(edit: a bit more words for clarity)

Nwakaego-Ego:
Hi @bryanhoffman, I am following up since it's been a while without a response here. @calebfoss left a couple of open questions above. Let us know if you are still working on this, and also if you need help. Thank you for your work on this so far.

bryanhoffman:
@Nwakaego-Ego Thanks for the comment! I have not tested to see if the fix works in instance mode like @calebfoss brought up. My first PR was a no-go for accessibility but I am hopeful for feedback on the addition of a supressScrolling function outlined here: https://github.com/processing/p5.js-website/pull/1278#issuecomment-4195162496

Issue #1248, opened 2026-03-16
build: update all github actions away from deprecated node.js 20


nbogie:
To ensure the continued smooth running of the site's CI systems, all used github actions that currently run on node 20 should be updated to versions which run on node 24.

### Why?

As per [this notice](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) from github blog, 
"Node20 will reach end-of-life (EOL) in April of 2026. As a result we have started the deprecation process of Node20 for GitHub Actions."

### More info: 

When website `build-deploy` job runs (for example) the following warning is issued:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-node@v4, actions/upload-artifact@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24.

### Testing

Properly resolving this issue will be 2% code change and 98% doing thorough testing of the affected CI flows.

### See also

#1233

### See also

There is a similar task to update the CI actions for the p5.js repo:
https://github.com/processing/p5.js/issues/8674

nbogie:
I'd like to work on this.

Update:
Moving setup_node and checkout actions to version 5 tests out fine.  However, beta_deploy gets a warning for [cloudflare/wrangler-action@v3](https://github.com/cloudflare/wrangler-action) running on node 20, and that action doesn't have a more up-to-date version.  

Wrangler [issue here](https://github.com/cloudflare/wrangler-action/issues/413) has some movement.

The warning also gives updated timelines:

> Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: cloudflare/wrangler-action@v3. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24.

ulgens:
@nbogie I checked #1282 but it looks like an automation solution is not a part of the PR (yet). If that sounds useful, I can help to add a Renovate (or dependabot) config with it, so manual work needed for similar updates can be minimized.

nbogie:
Hi, @ulgens!  That's something I'll leave the maintainers to comment on.  Perhaps @limzykenneth

limzykenneth:
If dependabot can do it that would be great as it already raises NPM dependency PRs for us.

[...]

PR #1282, opened 2026-03-25
Upgrade GitHub actions from node 20

nbogie:
addresses: #1248 

Changed beta_deploy.yml and test.yml and deploy.yml to use:
* actions/checkout v4 -> v5
* actions/setup-node v4 -> v5, or  (leave where already @v6.3.0)

Both of the above v5's are based on node 24, where the v4s were based on the soon-unsupported node 20.

### testing: 
* Tested `Test, typecheck, and lint`and all 4 jobs ran fine, with no warning  about action node versions.
* Tested build_deploy, which ran fine to cloudflare.   however this workflow got a warning as the `cloudflare/wrangler-action@v3` we are using in deployment is node-20 based.  There isn't an updated one, yet.

### DIFF
--- a/.github/workflows/beta_deploy.yml
+++ b/.github/workflows/beta_deploy.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           node-version: 22.22.x
       - name: Checkout your repository using git
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
       - name: Install dependencies
         run: npm ci
       - name: Build website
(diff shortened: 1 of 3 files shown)


ulgens:
I believe this is superseded and can be closed: https://github.com/processing/p5.js-website/issues/1248#issuecomment-4593916809

ksen0:
Thanks for bumping, closing.

PR #1289, opened 2026-03-31
fix createCapture preview clipping

### LINKED ISSUE #1204: The examples on the createCapture reference don't display well

aashu2006:
Resolves #1204 

Adds `Capture` to the DOM check so `createCapture()` examples get extra preview height. Fixes the issue where the video was getting cut off below the canvas. Follows the same logic used for other DOM functions.

### DIFF
--- a/src/components/CodeEmbed/index.jsx
+++ b/src/components/CodeEmbed/index.jsx
@@ -45,7 +45,7 @@ export const CodeEmbed = (props) => {
   }
 
   // Quick hack to make room for DOM that gets added below the canvas by default
-  const domMatch = /create(Button|Select|P|Div|Input|ColorPicker)/.exec(initialCode);
+  const domMatch = /create(Button|Select|P|Div|Input|ColorPicker|Capture)/.exec(initialCode);
   if (domMatch && previewHeight) {
     previewHeight += 100;
   }


aashu2006:
Hi @ksen0 . This was because `createCapture` wasn’t included in the DOM check, so the video was getting cut off below the canvas.

added it now and example 3 looks correct.

for the other examples:
- example 2 seems fine (video is hidden and drawn on canvas, just small)
- example 4 is `norender`, so not part of this

let me know if this looks good

doradocodes:
@aashu2006 While I was testing your branch against the `main` and `2.0` branches, I noticed that Example 4 on that page is working differently:

A local version of the `main` branch with your change:
Screenshot 2026-04-13 at 11 57 08 AM

A local version of the `2.0` branch with your change:
Screenshot 2026-04-13 at 12 06 20 PM

I'm wondering why the preview window appears next the the code snippet for `main` but not `2.0`. Is this something you've encountered?

aashu2006:
@doradocodes yeah I checked this.

The preview is rendering in both cases, but the layout is different:
- main shows it side-by-side
- 2.0 stacks it above

I didn’t touch anything related to layout or preview positioning, the change only affects the height for createCapture.
So this seems to be a difference in layout behavior between the branches, not caused by this PR.

Let me know if you want me to look into that separately.

Nwakaego-Ego:
Hi @doradocodes, Just flagging PR #1289 to your attention. Thank you.

doradocodes:
@aashu2006 I think the real issue is that the video preview for the last example is not showing on the `main` branch, which is an issue I also see on production. So I agree with you, the issue is unrelated to this PR, and we will make a new issue for that. 

Thanks for this fix, we will merge this now.

doradocodes (approved):

PR #1259, opened 2026-03-19
feat(libraries): replace p5.treegl with p5.tree

nakednous:
Replaces the deprecated p5.treegl (IIFE-only, v1) with p5.tree, its actively maintained successor.

- Supports both IIFE (CDN) and ESM (npm)
- Targets p5.js v2 (WEBGL / WEBGL2 / WebGPU)
- Expanded scope: render pipeline, pose/camera interpolation, 
  post-processing, frustum visibility, picking, HUD, 
  and declarative control panels
- Licensed AGPL-3.0
- Same author, same featured image

PR #1300, opened 2026-04-04
Replace p5.treegl with p5.tree in 2.0

nakednous:
Ports the already-merged library update from #1259 to `2.0` as requested, replacing deprecated `p5.treegl` with `p5.tree` for the beta site.


limzykenneth:
Looks good. Thanks!

PR #1284, opened 2026-03-28
fix: normalize p5 reference links in parameter sections

rushdarshan:
Fixes #674

The regex in normalizeP5ReferenceLinks() was capturing the 'p5' prefix in the capture group, causing duplicate 'p5' in URLs.

Before: #/p5/rectMode -> /reference/p5/p5/rectMode/ (broken)
After:  #/p5/rectMode -> /reference/p5/rectMode/ (fixed)

Updated regex to exclude 'p5' from capture group while still handling both dot notation (#/p5.Element) and slash notation (#/p5/rectMode) patterns.

Issue #674, opened 2025-01-07
Links in Parameters section don't work


ademuri:
### Most appropriate sections of the p5.js website?

Reference

### What is your operating system?

Linux

### Web browser and version

Google Chrome 131.0.6778.204 (Official Build) (64-bit)

### Actual Behavior

I believe that links in the Reference are auto-generated from the Javascript comments. Links in the description of the function work, but links in the Parameters section do not.

For example, on the [Text page](https://p5js.org/reference/p5/text/):
- First link to `rectMode()`, in the description, correctly links to: https://p5js.org/reference/p5/rectMode/
- Last link to `rectMode()`, in the Parameters, incorrectly links to: https://p5js.org/reference/p5/text/#/p5/rectMode. This causes nothing to happen (invalid anchor link).

### Expected Behavior

Links in the Parameters section should correctly link to other pages, as in the rest of the reference page.

### Would you like to work on the issue?

Sure, I can work on this if someone can point me to where this is configured or generated

wong-justin:
Similar to #676; this link slipped by the checker in #527

sushanthraiurwa:
Hi, I’ve reproduced the issue locally.

From investigating src/layouts/Reference.astro, it looks like the problem comes from normalizeP5ReferenceLinks().

Currently, the regex only handles anchors of the form #/p5.SomeClass, but links in the Parameters section are generated as #/p5/someFunction.

Because of that, those links aren’t being normalized to /reference/p5/.../, which causes the broken anchor behavior.

I’m thinking of updating the normalization logic to support both #/p5.ClassName and #/p5/functionName formats.

Please let me know if that approach sounds good.

rushdarshan:
Hi @ksen0 — is this still open for work? I can see @sushanthraiurwa identified the fix 5 weeks ago but no PR has been opened. Happy to implement the normalization update for both #/p5.ClassName and #/p5/functionName patterns if it's still available.

ksen0:
Hi @sushanthraiurwa @rushdarshan , sorry I missed this! Thanks for bumping it.

@sushanthraiurwa since you volunteered first, please feel free to claim this.
@rushdarshan if the previous volunteer doesn't reply to claim in 7 days, please feel free to work on it.

rushdarshan:
Hi @ksen0, since it's been 7 days, I'll go ahead and claim this! I'll start working on updating the normalization logic in src/layouts/Reference.astro and will open a PR shortly.

PR #1310, opened 2026-04-07
fix: enable OpenProcessing sketch embeds on detail pages

### LINKED ISSUE #1307: Sketch detail pages: OpenProcessing embeds stuck in loading state, never render

Community (Sketches, Libraries, Events)

None

### Web browser and version

_No response_

### Actual Behavior

On all sketch detail pages (`/sketches/{id}`), the embed area displays a loading 
spinner indefinitely. The sketch content never renders and no error message is shown.

Tested on:
- https://p5js.org/sketches/2690038/ ("Feelings make us human")
- https://p5js.org/sketches/2484739/ ("All you need is infinite love!")
- https://p5js.org/sketches/2689119/ ("Boiling Point")

All three reproduce the same failure consistently.

Image

### Expected Behavior

The embedded sketch should render in the embed area, or a graceful fallback 
should be shown if the embed fails to load.

### Steps to reproduce

1. Open any sketch detail page e.g. https://p5js.org/sketches/2690038/
2. Wait for the page to fully load
3. Observe the embed area — loading spinner appears and never resolves

Root cause: iframe elements are missing `sandbox` and `allow` attributes, 
which blocks OpenProcessing's embed scripts from executing.

Yes, I have a fix ready and can open a PR immediately.

rushdarshan:
## Description
 OpenProcessing sketch embeds on sketch detail pages were failing to render due to missing iframe security/permission attributes.
 
 ## Root Cause
 iframe elements lacked required \`sandbox\`/\`allow\` attributes, which could block embed scripts.
 
 ## Solution
 - Added sandbox and allow attributes for OpenProcessing embeds
 - Added lazy loading where applicable
 - Kept fallback iframe behavior aligned
 
 Resolves #1307


Nwakaego-Ego:
Hi @rushdarshan thank you for this PR. I noticed a few things while reviewing:

This PR contains two separate fixes. The reference link fix for issue #674 is already addressed in your separate PR #1284. Including it here creates duplication.
This PR is not linked to issue #1307. Please add "Resolves #1307" to the PR description so GitHub can track the connection.
I also checked the three sketch pages reported in issue #1307 and they are now loading correctly on the live site without this PR being merged. 

cc @ksen0 @doradocodes to advise

ksen0:
Thanks for the work @rushdarshan and the review @Nwakaego-Ego !

I've left a review on the other PR; if that one gets merged first, then the commit history of this wil also be cleaned up after updating the branch. But I would suggest in the future to work in separate local branches, or on one PR at a time, so that there's no extra commits.

If you're still interested in working on it, please complete the work on the other fix PR. The OpenProcessing loading problems I think have a different root cause, but it's really hard to test and verify. Still, the proposed fix here is helpful, but because of the extra commits it must be reviewed after the other open PR is done. (Or you can close this one and reopen with a cleaned-up history without extra commits.)

Thanks for your time and patience.

rushdarshan:
Thanks for the review! I cleaned this PR so it now only contains the OpenProcessing embed fix (removed the
 #674/reference-link changes) and updated the description with **Resolves #1307**. Could you please re-review?

Nwakaego-Ego:
Hi @rushdarshan, thank you for working on this.

I tested the three sketch pages you shared from issue #1307, and they are all loading correctly on the live site without this PR being merged, which may suggest the root cause was on OpenProcessing's side rather than the website code. I stand to be corrected on this.

The sandbox attribute is still a valid and worthwhile addition as good practice.

However, there are a few concerns:

The original issue only asked for the sketches to render. Adding allow="fullscreen; clipboard-write" and the error fallback UI goes beyond that scope. These are feature decisions that affect how the site behaves and should be discussed and approved through a separate issue before being added. 

Also, comparing your version with the original ScalingIframe component on main, the following are missing from your iframe:

left: 0
transform: scale(${scale.toFixed(4)})
transformOrigin: "top left"
ref={iframeRef}

These are what make the scaling work. 

cc @doradocodes @ksen0, kindly advise.

PR #1313, opened 2026-04-07
sass: Migrate from deprecated @import to @use

### LINKED ISSUE #1309: build: Fix noisy sass deprecation warnings by migrating away from @import
Running `npm run dev` outputs 300+ lines of deprecation warnings regarding sass `@import` rules.

These messages are noisy and distracting and will quite possibly cause developers to miss other more useful warnings, and cause new developers to worry that something is not set up correctly on their environments.

You can [read about the deprecation on the official blog post](https://sass-lang.com/blog/import-is-deprecated/).

### Example message:

> Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
> 
> More info and automated migrator: https://sass-lang.com/d/import
> 
```
  ╷
2 │   @import "../../styles/variables.scss";
  │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
```
> src/layouts/HomepageLayout.astro 2:11  root stylesheet

### Proposal

These messages [can be silenced](https://sass-lang.com/blog/import-is-deprecated/#controlling-deprecation-warnings), and the `@import` mechanism is not expected to be removed until Dart Sass 3.0.0 (["timeline"](https://sass-lang.com/blog/import-is-deprecated/#timeline)).

However, I suggest we migrate away from the use of `@import` now.  (This will give us time to solve any resulting teething problems, rather than eventually be time-pressured to rush migration and trouble-shooting in future.)

Include [config](https://sass-lang.com/blog/import-is-deprecated/#controlling-deprecation-warnings) so that future uses of `@import` break the build immediately, to make them easy to spot and prevent.

(I suggest we migrate on both  2.0 _and_ main, keeping them in sync and making it easier for future changes to be applied to both branches.)

### Reading

* [Import is deprecated](https://sass-lang.com/blog/import-is-deprecated/)
* [Differences between `@use` and `@import`](https://sass-lang.com/documentation/at-rules/use/#differences-from-import) 
* [Docs on why and how to migrate from `@import`](https://sass-lang.com/documentation/breaking-changes/import/)

nbogie:
Fixes #1309 , moving away from the deprecated `@import` to `@use`.  

### Changes:
* moves from use of `@import` to `@use`
* changes paths used in import/use directives, removing leading slash: 
e.g.  `/styles/variables.scss` to `styles/variables.scss`
* configures sass through astro's vite config so that future use of `@import` (an accidental regression) will break the build with appropriate message.

### Status: 
Done, I think. Ready for review.

[x] dev and build work locally and look fine.
[x] deploys to cloudflare through the normal github [actions](https://github.com/nbogie/p5.js-website/actions/runs/25281672157/job/74119803493): https://p5web-sass-fix-beta.neillzero.workers.dev/
[x] looks ok deployed this way
[x] I ran a comparison of generated css files before-and-after and found minimal differences (no additional css duplication, for example).
[x] No more deprecation warnings from this source in `npm run dev`.

#### All CSS files in dist/ - before and after

Locally ran astro build with `npm run build` on the current 2.0 with and without the PR, and compared all generated `.css` files.
* Result: same file sizes, slightly different names for about.whatever.css
* The all.css and _slug_.ZGkxnZFm.css contents are identical.
* The about.whatever.css differs only in ids and in section ordering.
```
 40K	./_astro/about.Cqki5N6P.css
 24K	./_astro/_slug_.ZGkxnZFm.css
 44K	./assets/all.css
```
A main worry was that the new approach would add repeats of css sections, bloating downloads.  No file size increase here is good to see.

However, I understand that astro can also inline css into html pages.  So i tested the generated sizes of those next.

#### dist/ dir - before and after
* The overall dist directory got _smaller_ from 386M to 381M.

#### All html files in dist/ - before and after

* The large majority of files didn't change significantly in size as reported by `du` (uses 512 byte blocks). 
* Looking at the english reference page sizes down to the byte, the large majority of these differ only in size only by a single digit of bytes.

#### CI including deployment
* Tested the "deploy beta to cloudflare" workflow on github CI, with a minimally-altered branch (to tweak wrangler.jsonc) - It built and deployed fine to cloudflare, and looks ok there: https://p5web-sass-fix-beta.neillzero.workers.dev/ .  ([github CI workflow logs here](https://github.com/nbogie/p5.js-website/actions/runs/25281672157/job/74119803493))

### discussion (opinions)

While it's possible to rename the namespace when using `@use`, 
So instead of:
```
@use "styles/variables.scss";
@media (min-width: variables.$breakpoint-tablet) {
```
We could write:
```
@use "styles/variables.scss as v";
@media (min-width: v.$breakpoint-tablet) {
```

but I suggest sticking with the full namespace (in this case `variables`) is less surprising, more self-descriptive, easier for find/find-replace, and not too long.

(Anyway, that can be decided and changed going fwd.)


limzykenneth:
Full namespace is fine as it is not very significant. Let me know when you are ready with this one and I can check locally as well.

nbogie:
Hi @limzykenneth! I've tested this out and detailed the **status** and **testing** sections in the top body. 
If you can test locally that'd be great, thank you!

limzykenneth:
All looks good. Ready for merge?

nbogie:
yes.  thank you for taking a look.

PR #1320, opened 2026-04-08
Update Node.js version from 18 to 22 in translation tracker

ksen0:
Workflow has been failing due to Node version.


Divyansh013 (approved):
Thanks for this @ksen0 !

Issue #1328, opened 2026-04-09
Tutorials: Conditionals and Interactivity/Organizing Code with Functions: Sunrise Animation or Sunset Animation?


Chiuye:
In the tutorials _Conditionals and Interactivity_ and _Organizing Code with Functions_, some places describe the animation as **sunrise animation**, while others call it **sunset animation**. Wouldn't it be better to unify it as **sunrise animation**?

<img width="1203" height="721" alt="Image" src="https://github.com/user-attachments/assets/a6f72db1-6148-4174-9bca-3033d36cc0d3" />

<img width="1167" height="533" alt="Image" src="https://github.com/user-attachments/assets/6c02775a-84cd-4107-a6e4-4d27dd7985f4" />

<img width="1133" height="788" alt="Image" src="https://github.com/user-attachments/assets/57e06579-bc2e-4cc8-af56-0ca9f0f107b8" />

Chiuye:
@ksen0 I‘d like to fix this~ >w<

Chiuye:
@ksen0 Here~~~~ Look at this~~~

PR #1349, opened 2026-04-21
Tutorials: unify the names to "sunrise animation"

Chiuye:
unify the names to "sunrise animation" in tutorials: Conditionals and Interactivity/Organizing Code with Functions

Addresses #1328


Chiuye:
Hi @ksen0 , I've created this PR to fix issue #1328 . These changes are both in English and Chinese. Please help review it~
For the Chinese modifications, @lirenjie95 may need to take a look as well? 😆🥳

Besides, I think we also need to commit these changes to 2.0 branch. But that is the last step right?

lirenjie95 (approved):
LGTM

ksen0:
Thanks for the fix and the patience!

PR #1377, opened 2026-04-28
Remove filter from reference page

### LINKED ISSUE #1364: Remove filter on reference page
Recent [usability testing](https://github.com/processing/p5.js-website/issues/1221) has identified this feature is very confusing. It is secondary to the site-wide search, and does not behave as expected (e.g., it does nto search descriptions)

screenshot of reference page with 'circle' typed into the filtering field

In the future, we would like to add tags/keywords that one could use buttons to filter through. However, the filtering mechanism can be added back in at that time. For now, filter functionality should be removed completely to improving website maintainability and clarity.

aashu2006:
**Description:**

Fixes #1364

This PR removes the keyword filter input from the reference page. 

**Changes:**
- Removed filter input bar, search state, and filter logic from `ReferenceDirectoryWithFilter`
- Removed `client:load` from layout, the component is now static
- Removed unused `Filter by keyword` and `No Results` translation keys from all locales too (en, es, hi, ko, zh-Hans)

### DIFF
--- a/src/components/PageHeader/RootPage.astro
+++ b/src/components/PageHeader/RootPage.astro
@@ -1,14 +1,7 @@
 ---
 const { title, subtitle } = Astro.props;
 
-// The page header and filter by keyword are separated in the hierarchy but
-// are intended to visually blend together. A bottom border should be added
-// to the header only when the filter by keyword is not present.
-const routesWithFilter = ["/reference"];
-const isFilterRoute = routesWithFilter.some((route) =>
-  Astro.url.pathname.includes(route)
-);
-const borderStyle = isFilterRoute ? "" : "border-b border-sidebar-type-color";
+const borderStyle = "border-b border-sidebar-type-color";
 ---
 
 <div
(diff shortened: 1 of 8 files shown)


ksen0:
Nice, thank you! @aashu2006 could you add a screenshot of after as well please? I would expect a big blank space now like the other top-level categories

aashu2006:
#### Before: filter bar visible
image

#### After: filter bar removed
image

aashu2006:
@ksen0, I also noticed that the `RootPageHeader.astro` had special logic to hide the bottom border on the reference page because of the filter bar. So I've fixed that so the border now shows consistently across all root pages. Let me know if that looks good!

Nwakaego-Ego:
Hi @aashu2006 Thank you for this PR. I tested it locally and can confirm:
The reference page now looks consistent with other top level pages. The filter bar is completely gone and the header flows cleanly into the content below. The translation keys were also cleanly removed from all 5 language files. Good catch on the border fix, too, the border now shows consistently across all pages. 

cc @ksen0 @doradocodes

PR #1382, opened 2026-04-29
Restore No Results translation string (used by search)

aashu2006:
Follows up on #1364.

Restores the No Results translation key in en, ko, and zh-Hans, it was accidentally removed as part of the filter removal, but it's also used by site-wide search, not just the filter.

### DIFF
--- a/src/content/ui/en.yaml
+++ b/src/content/ui/en.yaml
@@ -43,6 +43,7 @@ Related Examples: Related Examples
 Show Code: Show Code
 Donate to p5.js: Donate to p5.js
 Download p5.js: Download p5.js
+No Results: No entries found for that search.
 No alt text: No alt text
 briefPageDescriptions:
   Reference: Find easy explanations for every piece of p5.js code.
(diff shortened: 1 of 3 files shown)


aashu2006:
Hi @doradocodes @Nwakaego-Ego , please let me know with any suggestion or tweaks if needed. Thanks!

Nwakaego-Ego:
Thank you, @aashu2006. On the main branch, No Results string has been restored and is working correctly. On the 2.0 branch, PR #1383 applies all the same changes cleanly, filter bar removed, header border consistent, and No Results working as expected.

No Results across all languages for both branches:
English: No entries found for that search.
Chinese: 未找到相关条目
Korean: 검색 결과가 없습니다
Spanish: Falls back to English
Hindi: Falls back to English

image

image


I went back and checked PR #1377 to understand why Spanish and Hindi were falling back to English. Looking through the files that were changed, the No Results key was never in es.yaml or hi.yaml to begin with, so this is not something this PR introduced. I will open a separate issue for adding the Spanish and Hindi translations.

cc @doradocodes @ksen0

PR #1378, opened 2026-04-28
Update p5.waves featured image

seb-prjcts-be:
Replaces the existing thumbnail with an updated hero composition. Same dimensions (1500×1000) as the current image.

PR #1384, opened 2026-04-29
Update p5.waves featured image (2.0)

seb-prjcts-be:
Mirrors #1378 on the `2.0` branch as requested by @ksen0.

Replaces the existing thumbnail with the same updated hero composition (1500×1000), identical bytes to what was merged on `main` in #1378.


ksen0:
Thanks, sorry for the manual process!

Issue #1207, opened 2026-03-02
Korean 'Jump to' menu shows Chinese


picxenk:
### Most appropriate sections of the p5.js website?

Other (specify if possible)

### What is your operating system?

Mac OS

### Web browser and version

145.0.7632.109 (Official Build) (arm64) 

### Actual Behavior

Although the webpage language is set to Korean, the "Jump to" menu on all pages is displayed in Chinese characters.

### Expected Behavior

If the language is set to Korean, the text in the "Jump to" menu should also be displayed in Korean.

### Would you like to work on the issue?

No

lirenjie95:
It also happens when using Spanish (the jump to menu shows some Hindi links). The “Jump to” menu is driven by a module-level global (jumpToState) that is mutated during page rendering, which is unsafe under Astro’s concurrent SSR/static generation. When multiple locales build in parallel, one page can overwrite jumpToState while another page’s Nav reads it, causing cross-locale leakage (e.g., Korean pages showing Chinese labels). This is why it doesn’t reliably fall back to English—it's not a translation fallback, but a race condition from shared mutable state.

ksen0:
@lirenjie95 Thanks for the analysis! I've marked it as open for work (unless you want to claim it?)

lirenjie95:
@ksen0 please keep it open to everyone. I haven't figured out the solution yet.

abhi-2028:
@ksen0 Hey, 

I cloned the repository and tested the issue locally. On the deployed website, the “Jump to” menu shows Hindi/Chinese text even when the language is set to Spanish or Korean .

<img width="650" height="643" alt="Image" src="https://github.com/user-attachments/assets/6d9c1995-7c05-4d49-82cf-a3731702dd42" />

However, when I run the project locally after cloning the repo, the “Jump to” menu appears correctly in the selected language, and I cannot reproduce the cross-locale text issue.

<img width="641" height="750" alt="Image" src="https://github.com/user-attachments/assets/bc97c036-07df-41c6-8740-ff35b13bc735" />

Because of this difference between the production site and the local environment, I’m wondering:

Could this be related to Astro’s concurrent static generation during the production build rather than the dev environment?

Is there a specific build command or environment configuration used for the official site that might trigger the race condition mentioned earlier (jumpToState global mutation)?

Is there a recommended way to reproduce the production build locally so contributors can properly debug the issue?

Happy to investigate further once I can reproduce the problem reliably. Thanks!

[...]

PR #1391, opened 2026-05-06
Fix Wrong Page Underlined as Current Page in Left Nav Bar

### LINKED ISSUE #1304: Incorrect information in detailed left side navigation styling

Home

Mac OS

### Web browser and version

(Any)

### Actual Behavior

Navigate to [an example](https://p5js.org/examples/calculating-values-interpolate/) or [a tutorial](https://p5js.org/tutorials/variables-and-change/) and observe that on the lower left, it appears that the current page is underlined. However, it is usually incorrect.

### Expected Behavior

The underlined name of tutorial or example should match the current tutorial or example.

### Steps to reproduce

* [an example](https://p5js.org/examples/calculating-values-interpolate/) should have "Linear Interpolation" underlined on the left, but has "Map"
* [a tutorial](https://p5js.org/tutorials/variables-and-change/) should have "Variable an Change" underlined on the left, but has something in another section

(reserved for CodeDay microinternship)

esu-skoopin:
Resolves #1304 

### Perceived Cause Behind Bug

The main issue, as far as we know, is the combination of a build configuration that allows for the site to be built concurrently two pages at a time and the use of the global variable, `jumpToState`.

<br>

This is the relevant build setting:
```
// Code location: astro.config.mjs

build: {
  format: "directory",
  concurrency: 2
},
```
<br>

We can take a simulated look at what's happening to better understand the problem:

| Step | What Happens                                                                | Page Associated with `jumpToState` |
|------|-----------------------------------------------------------------------------|------------------------------------|
| 0    | Pages A and B are being built at the same time | — <tr></tr>                                  |
| 1    | Set `jumpToState` for page A                          | A <tr></tr>             |
| 2    | Set `jumpToState` for page B                                            | B <tr></tr>                              |
| 3    | Read from `jumpToState` for page A     | B <tr></tr>                              |
| 4    | Read from `jumpToState` for page B                      | B <tr></tr>                              |

<br>

Because the site is being built concurrently two pages at a time, it's possible for `jumpToState` to be overwritten by page B before page A can read from `jumpToState`.

<br>

This can result in the visual bug, as seen below.

<table>
  <tr>
    <td>Screenshot of 'Linear Interpolation' example page, showcasing visual bug in left nav bar<br><i>Current page incorrectly underlined as "Map"</i></td>
    <td>Screenshot of 'Map' example page<br><i>Current page correctly underlined as "Map"</i></td>
  </tr>
</table>

<br>

*The assumption is that these pages were built concurrently at build time and "Map" overwrote `jumpToState` before "Linear Interpolation" could read from it

<br>

### Our Solution

For our solution, we didn't want to simply change the build config for the site, as we believe that this config setting was most likely put in place for a good reason, so we instead decided to replace the use of the global `jumpToState` variable with a "jumpToState" prop for the &lt;BaseLayout&gt; and &lt;Nav&gt; components.

<br>

| Old Approach                                                                                                                                                                             | New Approach                                                                   |
|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------:|
| Set global `jumpToState` variable<br> ↓<br> Global `jumpToState` variable potentially gets overwritten by another page<br> ↓<br> &lt;Nav jumpToState={global `jumpToState` variable}&gt; | &lt;BaseLayout jumpToState={page}&gt;<br> ↓<br> &lt;Nav jumpToState={page}&gt; |

<br>

#### Screenshots of our solution working as expected locally:

- "Linear Interpolation" page
Screenshot of local instance of 'Linear Interpolation' example page

&nbsp;<br>

- "Variables and Change" page
Screenshot of local instance of 'Variables and Change' tutorial page

&nbsp;

Feedback welcome!


Nwakaego-Ego:
Hi @esu-skoopin thank you for this PR. I tested locally and can confirm the fix works correctly. 
cc @ksen0 @doradocodes

esu-skoopin:
@lirenjie95  Would you happen to be able to review this PR? I believe the changes I've made for this PR are very similar to PR #1271 which you reviewed a month ago.

lirenjie95 (approved):
This is a cleaner and more robust approach than #1271—eliminating the shared mutable state entirely instead of working around it. I also believe this will resolve #1207, since the cross-locale "Jump to" leakage was caused by the same jumpToState race condition. One more thing, please squash and merge to keep the commit history tidy. Thanks for the great work!

ksen0:
Thanks for the fantastic work @JazzJE @RyRy241 @esu-skoopin, and big thanks to the reviewers @Nwakaego-Ego @lirenjie95 ! Merging this one

PR #1392, opened 2026-05-07
Improve Korean typesetting

### LINKED ISSUE #1376: Improve line spacing for Korean documents
I’d like to suggest increasing the line spacing for Korean text across the website.

### Current behavior

The current line spacing for Korean content feels too tight. In many cases, the vertical spacing between lines is similar to the horizontal spacing between words, which makes it harder for the eye to naturally track lines of text.

Image
  
  
This issue is more noticeable in Korean, probably because of the glyph difference. Some fonts allocate less vertical space for Korean glyphs (because Korean characters don't have descenders like latin alphabet), which can make the default line height feel visually compressed.

### Suggested improvement

Increasing the line spacing (line height) would improve readability. For example:

Image
  
  
This can be done by adding following CSS rule to `styles/markdown.scss`

```scss
// Korean text needs more line spacing for readability
html[lang="ko"] .rendered-markdown {
  line-height: 1.8;
}
```

eupthere:
Addresses Issue #1376 

#### Added Korean line spacing settings for improved readability

image

#### Added `word-break: keep-all` to prevent awkward Korean line breaks

image
<br/>

Latin text has natural break points at spaces, and browsers preserve alphabetic words by default. Korean text, under default East Asian line-breaking rules, may allow breaks between syllable blocks. `word-break: keep-all` tells the browser to avoid those intra-word CJK breaks and keep Korean word groups together, more like Latin words.

### DIFF
--- a/styles/markdown.scss
+++ b/styles/markdown.scss
@@ -107,3 +107,8 @@
     padding-bottom: 0;
   }
 }
+
+// Korean text needs more line spacing for readability
+html[lang="ko"] .rendered-markdown {
+  line-height: 1.7;
+}
--- a/styles/markdown.scss
+++ b/styles/markdown.scss
@@ -108,7 +108,10 @@
   }
 }
 
-// Korean text needs more line spacing for readability
+
 html[lang="ko"] .rendered-markdown {
+  // Korean text needs more line spacing for readability
   line-height: 1.7;
+  // Prevent awkward Korean line breaks
+  word-break: keep-all;
 }


Nwakaego-Ego:
Hi @eupthere, thank you for this PR. I tested it locally on the tutorial and example pages and can confirm that the line spacing was applied correctly.
cc @doradocodes @ksen0

ksen0:
Thanks @Nwakaego-Ego @eupthere ! If one of you makes the matchign 2.0 PR I'm happy to merge also

PR #1403, opened 2026-05-19
add MDX heading permalinks for tutorial pages

### LINKED ISSUE #1398: Add permalinks for headings in tutorials
### Increasing Access

Our tutorials are fairly long and cover a lot of things, so it's helpful to be able to send someone a link directly scrolled to a section in the middle of the tutorial. 

Home

### Feature request details

See for example how MDN lets you do this: when you hover over a heading, you'll notice it's a link. Clicking it scrolls to that heading and changes the URL (e.g. https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation#examples) which can then be copied and sent to people.

Under the hood this means either adding an `<a>` tag to each heading with a unique `id` attribute, ideally uniquely made from the text of the heading. Will also need some CSS styling to make this discoverable but not overwhelming (we should probably not style headings like regular links.)

syam-bukkuru:
## Summary
Adds MDX heading permalinks for tutorial pages using component overrides in `TutorialLayout.astro`.

Closes #1398 

## Changes
- Introduced reusable `HeadingPermalink` component (Preact)
- Overrode `h1–h6` in tutorial MDX rendering only
- Preserved Astro-generated heading IDs
- Made entire heading clickable (links to `#fragment`)
- Added improved hover UX:
  - `#` indicator appears on hover/focus (left side, subtle color)
  - heading text underlines on hover
- Scoped styles under `.tutorials .rendered-markdown`
- Avoided nested anchor issues

## Demo
Short demo showing heading hover, permalink visibility, and fragment navigation:  
https://github.com/user-attachments/assets/ab1f6c54-a9a5-423d-860e-360aec349cd9

## Testing
- `npm run check` passes (no errors)
- `npm run build` succeeds
- Verified:
  - hovering a heading shows `#` indicator
  - clicking heading updates URL fragment
  - direct `#fragment` navigation scrolls correctly
  - headings without IDs do not break
  - headings with inline links do not create nested anchors

## Notes
- Service worker 404 seen in dev console is unrelated to this change
- No changes to routes, schemas, or global markdown behavior

### DIFF
--- a/src/components/Dropdown/styles.module.scss
+++ b/src/components/Dropdown/styles.module.scss
@@ -1,4 +1,4 @@
-@import "/styles/variables.scss";
+@use "/styles/variables" as *;
 
 .container {
   position: relative;
--- a/src/components/HeadingPermalink/index.tsx
+++ b/src/components/HeadingPermalink/index.tsx
@@ -1,0 +1,62 @@
+import type { ComponentChildren, HTMLAttributes, VNode } from "preact";
+
+type HeadingTag = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+
+type HeadingProps = HTMLAttributes<HTMLHeadingElement> & {
+  as: HeadingTag;
+  children?: ComponentChildren;
+  id?: string;
+};
+
+type HeadingOverrideProps = Omit<HeadingProps, "as">;
+
+const joinClasses = (...classes: unknown[]) =>
+  classes
+    .filter((value): value is string => typeof value === "string")
+    .join(" ");
+
+export const HeadingPermalink = ({
+  as: Tag,
+  children,
+  class: classProp,
+  className,
+  id,
+  ...props
+}: HeadingProps) => {
+  const hasPermalink = typeof id === "string" && id.length > 0;
+
+  const mergedClassName = joinClasses(
+    "heading-permalink",
+    hasPermalink && "heading-permalink--enabled",
+    classProp,
+    className,
+  );
+
+  return (
+    <Tag {...props} id={id} class={mergedClassName}>
+      {hasPermalink ? (
+        <a href={`#${id}`} class="heading-permalink__link">
+          {children}
+        </a>
+      ) : (
+        children
+      )}
+    </Tag>
+  );
+};
+
+const createHeadingOverride = (as: HeadingTag) => {
+  const Component = (props: HeadingOverrideProps): VNode => (
+    <HeadingPermalink {...props} as={as} />
+  );
+
+  Component.displayName = `HeadingPermalink${as.toUpperCase()}`;
+  return Component;
+};
+
+export const HeadingPermalinkH1 = createHeadingOverride("h1");
+export const HeadingPermalinkH2 = createHeadingOverride("h2");
+export const HeadingPermalinkH3 = createHeadingOverride("h3");
+export const HeadingPermalinkH4 = createHeadingOverride("h4");
+export const HeadingPermalinkH5 = createHeadingOverride("h5");
+export const HeadingPermalinkH6 = createHeadingOverride("h6");
--- a/src/components/HeadingPermalink/index.tsx
+++ b/src/components/HeadingPermalink/index.tsx
@@ -15,6 +15,48 @@ const joinClasses = (...classes: unknown[]) =>
     .filter((value): value is string => typeof value === "string")
     .join(" ");
 
+const isVNode = (child: ComponentChildren): child is VNode =>
+  typeof child === "object" && child !== null && "type" in child;
+
+const childrenContainAnchor = (children: ComponentChildren): boolean => {
+  if (Array.isArray(children)) {
+    return children.some(childrenContainAnchor);
+  }
+
+  if (!isVNode(children)) {
+    return false;
+  }
+
+  if (children.type === "a") {
+    return true;
+  }
+
+  return childrenContainAnchor(children.props?.children);
+};
+
+const getChildText = (children: ComponentChildren): string => {
+  if (Array.isArray(children)) {
+    return children.map(getChildText).join("").trim();
+  }
+
+  if (typeof children === "string" || typeof children === "number") {
+    return String(children);
+  }
+
+  if (!isVNode(children)) {
+    return "";
+  }
+
+  if (children.type === "img") {
+    const imgProps = children.props as { alt?: unknown };
+    if (typeof imgProps.alt === "string") {
+      return imgProps.alt;
+    }
+  }
+
+  return getChildText(children.props?.children);
+};
+
 export const HeadingPermalink = ({
   as: Tag,
   children,
@@ -24,22 +66,32 @@ export const HeadingPermalink = ({
   ...props
 }: HeadingProps) => {
   const hasPermalink = typeof id === "string" && id.length > 0;
+  const hasNestedAnchor = childrenContainAnchor(children);
+  const headingText = getChildText(children);
 
   const mergedClassName = joinClasses(
     "heading-permalink",
     hasPermalink && "heading-permalink--enabled",
+    hasPermalink && hasNestedAnchor && "heading-permalink--with-marker",
     classProp,
     className,
   );
 
   return (
     <Tag {...props} id={id} class={mergedClassName}>
-      {hasPermalink ? (
+      {hasPermalink && !hasNestedAnchor ? (
         <a href={`#${id}`} class="heading-permalink__link">
           {children}
         </a>
       ) : (
         children
+      )}
+      {hasPermalink && hasNestedAnchor && (
+        <a href={`#${id}`} class="heading-permalink__marker">
+          <span class="sr-only">
+            {headingText ? `Permalink to ${headingText}` : "Permalink"}
+          </span>
+        </a>
       )}
     </Tag>
   );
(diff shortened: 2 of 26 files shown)


src/components/Dropdown/styles.module.scss
```diff
@@ -1,4 +1,4 @@
-@import "/styles/variables.scss";
+@use "/styles/variables" as *;
```
davepagurek, line 1: Was this to fix anything in particular? Ideally this change should not need to affect how we import other files, or update the package lock.

src/components/HeadingPermalink/index.tsx
```diff
@@ -0,0 +1,67 @@
+import type { ComponentChildren, HTMLAttributes, VNode } from "preact";
+
+type HeadingTag = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+
+type HeadingProps = HTMLAttributes<HTMLHeadingElement> & {
+  as: HeadingTag;
+  children?: ComponentChildren;
+  id?: string;
```
davepagurek, line 8: For my understanding, what passes this in? Since it seems to be implicit, we may want to add a comment to explain it.

src/components/HeadingPermalink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <Tag {...props} id={id} class={mergedClassName}>
+      {hasPermalink ? (
+        <a
+          href={`#${id}`}
+          class="heading-permalink__link"
+          aria-label="Link to this section"
+        >
+          <span class="heading-permalink__text">{children}</span>
+          <span class="heading-permalink__icon">#</span>
```
davepagurek: Not huge, but possibly this could be done with just one `<a>` tag as before without `<span>` children if we add the `#` via CSS? something like
```css
.heading-permalink::after: {
  content: '#';
  display: inline;
  /* ...etc */
}
```

src/layouts/TutorialLayout.astro
```diff
@@ -82,6 +90,12 @@ const { showBanner, englishUrl } = checkTranslationBanner(
         img: FreeRatioImage,
         pre: PreProxy,
         a: LinkWrapper,
+        h1: HeadingPermalinkH1,
```
davepagurek: While the tutorial layout is the main thing I had in mind when raising the issue, are there other pages that would benefit from this too? e.g. probably pages with the contributor docs should also have it. Put another way, should this be the default, maybe with other pages opting out if need be? @ksen0 let me know if you have thoughts there!

davepagurek:
Thanks for working on this, your screen recording looks like what I was imagining for this feature! Left some comments, let me know what you think.

src/components/HeadingPermalink/index.tsx
syam-bukkuru: MDX passes these props automatically when rendering headings. Astro already injects things like **id, children,** etc. into the heading components, so this just forwards them into our custom component.

src/components/Dropdown/styles.module.scss
syam-bukkuru: Those changes were to address local Sass deprecation warnings about @import being phased out in favor of @use. I updated them while testing.

src/components/HeadingPermalink/index.tsx
syam-bukkuru: yeah I thought about that, but since we’re making the whole heading a link, if the heading text contains inline links it could create nested anchors. that’s why I split it out earlier.

but yeah using ::after for # still makes sense.

src/layouts/TutorialLayout.astro
syam-bukkuru: I scoped it to tutorials first since that’s what the issue mentioned, but this could definitely be useful for other MDX pages like contributor docs too.

I wasn’t sure if it should be global by default or opt-in per layout, so I kept it limited for now. happy to extend it or move it to a shared layout if that’s preferred.

ksen0:
Thanks for working on this! Please link the issue in the PR description

syam-bukkuru:
> Thanks for working on this! Please link the issue in the PR description

Added the Issue Link in PR description. Please check once.

src/components/Dropdown/styles.module.scss
davepagurek: good to know, thanks!

src/layouts/TutorialLayout.astro
davepagurek: Makes sense, we can keep it limited at first, but I think at the very least let's do this in the contributor docs pages too to start with.

src/components/HeadingPermalink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    className,
+  );
+
+  return (
+    <Tag {...props} id={id} class={mergedClassName}>
+      {hasPermalink ? (
+        <a
+          href={`#${id}`}
+          class="heading-permalink__link"
+          aria-label="Link to this section"
```
coseeian: I suggest removing this aria-label so the browser can determine the accessible name from the actual link content.

Currently, generic text like "Link to this section" will repeat for every heading link, making them indistinguishable and confusing for screen reader users.

src/layouts/TutorialLayout.astro
syam-bukkuru: https://github.com/user-attachments/assets/935202a1-ff6d-428c-b63a-1d3f8f5bf768

I’ve extended the heading permalink behavior to contributor docs as well, while keeping it opt-in per layout rather than making it global for all markdown pages.

Tutorials and contributor docs now both use the shared heading permalink component, and the existing contributor docs jump-to links still use Astro’s generated heading slugs. I kept other MDX pages unchanged for now so we can expand deliberately if more pages need it later.

src/components/HeadingPermalink/index.tsx
syam-bukkuru: I am glad for your suggestion❤️. Made changes according to that, have a look into that.

coseeian: Looks good to me. Thank you.

styles/markdown.scss
```diff
@@ -108,10 +108,48 @@
...
+  }
+
+  .heading-permalink__link {
+    display: inline;
+    color: inherit;
+    text-decoration: none;
+  }
+
+  .heading-permalink__link::after {
+    content: "#";
```
coseeian: I suggest using alternative text syntax for the CSS-generated content to specify an empty string. This ensures screen readers do not read the character aloud:

```css
.heading-permalink__link::after {
  content: "#" / "";
}
```

Browser support is strong across modern engines, excluding legacy IE (see [Can I Use](https://caniuse.com/?search=content+alternative)). You can verify this by inspecting the accessible name in the browser DevTools Accessibility tab to ensure the "#" symbol is excluded from the calculated link name.

Image

For more details, see Sara Soueidan's article: https://www.sarasoueidan.com/blog/alt-text-for-css-generated-content/

syam-bukkuru: image

Made the changes.

src/layouts/TutorialLayout.astro
syam-bukkuru: https://github.com/user-attachments/assets/104a1410-b01c-45be-aa7b-4c8bb2915610

Applied permalink behavior for all pages, check once.

styles/markdown.scss
coseeian: Looks good to me. Thank you.

coseeian (approved):

syam-bukkuru:
Hi @ksen0 , @davepagurek , @coseeian once check these changes. If there are any modifications, please mention. else,  I like to clear the merge conflicts 😊.

ksen0:
HI @syam-bukkuru thanks for working on this, I really like how clear the indication is that the heading is linked! I also appreciate the sensible behavior of the links inside headings: both anchor link is visible and the additional link in the heading.

I did spot one side effect from generalizing this though:

1. Navigate to "Reference" > Typography section http://localhost:4321/reference/#Typography
2. Click on "Attributes"
3. Observe incorrectly being taken to an earlier sections' attributes subsection

Unless this is very straightforward to fix I'd suggest rather limiting the application of it on the Reference page to only h1/h2? cc @ritiupa, optionally, please feel free to weigh in on this.

Thanks to everyone for the careful reviews as well.

syam-bukkuru:
https://github.com/user-attachments/assets/05c8d54e-c213-4a5a-8c47-1cf85a82521f

Thanks for spotting that @ksen0 . Hi @ritiupa , @davepagurek , @coseeian,  I’ve updated the Reference page anchors so repeated subheadings are now context-aware and unique, e.g. #typography-attributes instead of a shared #attributes. That keeps the linked heading UX intact while avoiding the jump-to-the-wrong-section issue.  
I also updated the Reference jump-to links and the backlink from individual reference pages so they all use the same unique fragment format. So we shouldn’t need to limit this to just h1/h2 on Reference anymore.

ksen0:
Hi @syam-bukkuru ! Sorry for the delay, this looks great. Will you fix the merge conflicts as well? Please let us know in the next week or two; if you don't have time anymore we can also finish this up. Great work!

syam-bukkuru:
Hi @ksen0, I am occupied with work. Hope you wrap this up. Thank you, I had a great time working on our project ☺️.

PR #1409, opened 2026-05-25
[1.x] Major dependencies update

limzykenneth:
Update most dependencies and a major dependency update to Astro to v6. There are differences in Astro v5 and v6 which should have all been accounted for given the build is successful, we should keep an eye out for more.

Also includes PR continuous deployment preview with Cloudflare Workers versions. It currently only uploads the version, I'll need to update p5.js-comment-bot for it to also leave a comment here.

I'll cherry pick this PR to 2.0 branch once we can verify it works without issue.

I'm thinking of switching from ESLint/Prettier to oxlint/oxfmt as well but I'll probably defer that to another time.


limzykenneth:
Preview URL for current build is here: https://1e93bb25-p5-js-website.p5js-org-account.workers.dev/.

I'll need to rework the idea for PR preview.

ksen0 (approved):

PR #1421, opened 2026-05-31
[2.0] Fix reference URL not matching case sensitively

limzykenneth:
The URL should now match both original case of the reference file and all lower case instance of the URL. Note that it is not _case insensitive_, `reference/p5/creAteVecTor/` will not match, only `reference/p5/createVector/` and `reference/p5/createvector/`. 

_Except_, in preview mode on Mac (case insensitive file system, running `npm run preview`) which will then be fully case insensitive. This is due to the Astro's dev server and the case insensitive nature of the file system, it does not reflect actual production environment (which is always case sensitive), but also does not really have a meaningful impact. Just want to leave a full note here in case anyone stumble across this difference in behavior and got confused.


limzykenneth:
https://53cb4827-p5-js-website-beta.p5js-org-account.workers.dev

ksen0 (approved):

PR #1431, opened 2026-06-05
fix: Fixed scrollbar in example output frame (cherry-pick from #745)

### LINKED ISSUE #716: Scrolls in example code output frame seems unnecessary

Nwakaego-Ego:
Fixes #716

This is a clean cherry-pick of the relevant fix from #745 by @Its-sunny69 whose branch had carried over unrelated commits.

The fix removes the hardcoded width and height from the iframe in src/components/CodeEmbed/frame.tsx and lets it fill its parent container instead. Before this change, both the parent and the iframe were declaring the same size separately. Now, only the parent controls the size, which removes the confusion that was likely causing Chrome to show unnecessary scrollbars at certain zoom levels.

Note: I could not reproduce the bug locally or on the live site. I also found that index.jsx line 94 already has overflow-hidden on the parent container, which may explain why the bug is hard to reproduce consistently. The fix is still worth merging as it makes the code cleaner.

### DIFF
--- a/src/components/CodeEmbed/frame.tsx
+++ b/src/components/CodeEmbed/frame.tsx
@@ -165,9 +165,8 @@ export const CodeFrame = (props: CodeFrameProps) => {
         sandbox="allow-scripts allow-popups allow-modals allow-forms allow-same-origin"
         aria-label="Code Preview"
         title="Code Preview"
-        height={props.height}
-        width={props.width}
         loading={props.lazyLoad ? "lazy" : "eager"}
+        style={{width: "100%", height: "100%"}}
       />
     </div>
   );


doradocodes:
@Nwakaego-Ego Since this component is used often, can you double check this change on some of the other pages? 
For instance:
https://p5js.org/examples/animation-and-variables-mobile-device-movement/ <- should have a scrollbar on the first example

Nwakaego-Ego:
> @Nwakaego-Ego Since this component is used often, can you double check this change on some of the other pages? For instance: https://p5js.org/examples/animation-and-variables-mobile-device-movement/ <- should have a scrollbar on the first example

Hi @doradocodes, I checked the following locally

examples/animation-and-variables-mobile-device-movement/ scrollbar is still present  
examples/animation-and-variables-drawing-lines/ no  scrollbar 
examples/shapes-and-color-color-wheel/ no scrollbar
examples/animation-and-variables-animation-with-events/ no scrollbar 

Only the Mobile Device Movement example has a scrollbar. All other examples display without scrollbars.

doradocodes:
@Nwakaego-Ego Looks good, let's also cherry-pick this fix into `main`

Issue #1435, opened 2026-06-09
Intro to Shaders contains Broken Vertical text


RandomGamingDev:
### Most appropriate sections of the p5.js website?

Tutorials

### What is your operating system?

Mac OS

### Web browser and version

https://p5js.org/tutorials/intro-to-shaders/

### Actual Behavior

Completely vertical text that unnecessarily stretches the code boxes and is hard to read.

<img width="1710" height="982" alt="Image" src="https://github.com/user-attachments/assets/cb3afc0c-13f3-4855-b5be-6112aa0be443" />
<img width="1710" height="982" alt="Image" src="https://github.com/user-attachments/assets/7875dfc8-1145-4321-acb0-3b7b8933bbdb" />

### Expected Behavior

Formatting that allows the text to reach the ride of the screen until it hits padding in order to allow for readability.

### Steps to reproduce

Enter the site and scroll downwards to the glitched portion.

### Would you like to work on the issue?

Sure but it's fine if someone can fix it better since it's just a few lines

ksen0:
Also affects 2.x https://beta.p5js.org/tutorials/intro-to-glsl/

ksen0:
@RandomGamingDev if you have capacity an can fix, that would be great! Otherwise would @doradocodes @Nwakaego-Ego one of you be able to take a look please?

PR #1443, opened 2026-06-13
Fixed Intro to Shaders contains Broken Vertical text (#1435)

RandomGamingDev:
Fixed Intro to Shaders contains Broken Vertical text (#1435)


RandomGamingDev:
@ksen0

ksen0:
Thanks so much, confirmed works. (Still not perfect on mobile but definitely an improvement.)

PR #1460, opened 2026-06-20
Fix framebuffer blur vertex shader for p5.js 2.0

vamshi567bh:
Updates the Framebuffer Blur example vertex shader to use separate uModelMatrix + uViewMatrix uniforms instead of the deprecated uModelViewMatrix uniform. This fixes the incorrect matrix multiplication order that caused crescent-shaped distortion.

Fixes #8940


ksen0:
This should not be necessary, the GLSL code works on p5.js 1.x. The reason the example is broken is because of how the library has changed in 2.x, which was not intentional. I understand the approach and in another context it could apply, but in this case it misses the purpose of https://github.com/processing/p5.js-website/issues/1571 which is to address a regression from 1.x to 2.x. Sorry, and thanks for understanding!

Issue #1426, opened 2026-06-03
Switch websites so that v2.x deploys to `p5js.org` (the `beta.p5js.org` links continue to work via redirects)


ksen0:
[Please see the parent issue for timeline and context](https://github.com/processing/p5.js/issues/8870). lease feel free to raise questions, concerns, or overlooked tasks; however, the work will be done by the maintainers, so it is not open to implementation volunteers. However, additional non-blocking but very helpful tasks are at the bottom of the issue - volunteers are welcome to help!

When this task is done, on the `p5.js-website` repository, the `main` branch will include p5.js version 2 latest (currently `2.0`), and a `v1` branch will contain version 1 latest (currently `main`). Also, the `p5js.org` website will serve the main version 2.x reference, with `beta.p5js.org` redirecting to it to preserve existing URL usage. The version 1.x reference will remain accessible on `v1.p5js.org`.  This task affects `p5.js workflows` but not the editor, please see parent issue for more info.

## Steps maintainers will take
* Merge prerequisite PR that: 
  * Update links in `p5.js` repo on FES (including internationalization/localization strings) to correct URLs used by `p5.js#v1` to use [v1.p5js.org](http://v1.p5js.org/) for v1. No matching change is needed for v2. 
  * Updates contributor docs to use correct website links - mostly, should not be needed, and most links should be relative
  * There should be no links that become dead/bad after switch. Even better if there are NO external links (only to *.[p5js.org](http://p5js.org/))
  * updates all release workflows to refer to the correct branch
  * Ensure there are no .todo mdx (tutorials that are skipped for beta site) in `2.0` branch (in progress)
* Rename main to “v1?” and deploy it to v1.p5js.org
* Rename 2.0 to main branch and deploy
* Set up redirect for beta.p5js.org → [p5js.org](http://p5js.org/) to preserve old urls

## Tasks open to volunteers
If a volunteer would like to audit FES links of **external content** (outside of p5js.org), this would be really helpful! Every time there is Content that is only in a blog or p5.js wiki (not maintained) or mdn, but is important, it  should be added in the p5js.org website, and linked to correctly. However, this is not blocking for the switch, and even if a PR is made before the switch, it will likely not be merged until after.

We would also be grateful to volunteers for testing if existing community libraries listed on the website are working with 2.x. An audit was already done by maintainers when we released in April 2025, with proactive communication to the add-on library authors; and we are revisiting this audit again. However, more hands on deck is welcome 💟 

## Related 

[...]

PR #1469, opened 2026-06-26
Add scrollable and loadable from file EditableSketches

ksen0:
Currently the 2.0 switch is blocked by some .todo tutorials. These are blocked by incomplete migration, sometimes due to the presence of external links to sketches that cannot be updated to match 2.0. Currently they can't be copied directly into the body of the tutorial because they are very long. This PR introduces scrollable editable sketch support.

Addresses https://github.com/processing/p5.js-website/issues/1426 https://github.com/processing/p5.js/issues/8870

# WIP Docs

1. Put the .js/.css/.html into `.../tutorials/sketches/TUTORIAL_NAME/SKETCH_NAME`
2. use `export` at top of tutorial `.mdx` file to load the data

```js
//at the top
import { readFile } from "fs/promises";

export const completeSketchCode = await readFile(
  "src/content/tutorials/sketches/TUTORIAL_NAME/SKETCH_NAME/sketch.js",
  "utf-8"
);
export const cssCode = await readFile(
  "src/content/tutorials/sketches/TUTORIAL_NAME/SKETCH_NAME/main.css",
  "utf-8"
);

// etc for bodyCode for index.html
...

// in the doc body - sketch with own CSS+HTML
<SketchEmbed
  height="800px"
  code={completeSketchCode}
  cssCode={cssCode}
  bodyCode={bodyCode}
/>

// alternatively - editable sketch
<EditableSketch code={completeSketchCode} />

// editable sketch with own CSS+HTML may have some unexpected ideas about width.
// however explicitly setting width will use "vertical" mode where the code floats on the right.
<SketchEmbed
  height="800px"
  scrollable={true}
  code={completeSketchCode}
  cssCode={cssCode}
  bodyCode={bodyCode}
/>

```


src/components/CodeEmbed/index.jsx
```diff
@@ -125,11 +138,12 @@ export const CodeEmbed = (props) => {
           </div>
         </div>
       ) : null}
-      <div className="code-editor-container relative w-full">
+      <div ref={editorRef} className="code-editor-container relative w-full">
         <CodeMirror
           value={codeString}
           theme="light"
           width="100%"
+          height="300px"
```
davepagurek: Do we want to make this optional/opt-in for long sketches?

ksen0: Ah yes thanks for the reminder! That's the intent. (For other testers/reviewers looking later: when I add missing tutorials they shouldn't have anything changed except some scrollable sketches where intended.)

aashu2006 (commented):
hi @ksen0 , I tested locally with a long WEBGL sketch. The scrollable editor works well: code caps at 300px and scrolls, the page layout stays fine, edit and run works, and leaving out `scrollable` also changes nothing

also I feel there are two things to fix:

**1. Fold gutter shows on every editor now** (`CodeEmbed/index.jsx:160`). `foldGutter: true` is not gated on `scrollable`, so all tutorial and reference pages get the ▾ arrows

```diff
-            foldGutter: true,
+            foldGutter: isScrollable,
```

**2. `cssCode` / `bodyCode` never reach the sketch** ([`CodeEmbed/index.jsx:109-121`](https://github.com/processing/p5.js-website/blob/11966beb26d0965637f89001732eea96249cb436/src/components/CodeEmbed/index.jsx#L109-L121)). `EditableSketch` passes them to `CodeEmbed`, but `CodeEmbed` doesn't forward them to `CodeFrame`, so the CSS+HTML example loses both

```diff
             <CodeFrame
               jsCode={previewCodeString}
+              cssCode={props.cssCode}
+              htmlBodyCode={props.htmlBodyCode}
```

I tried both fixes locally, and untouched pages match `main` again after the gutter fix

ksen0:
Thank you for catching that @aashu2006 , added.

src/components/CodeEmbed/index.jsx
skyyash: currently the height prop of `EditableSketch` component only resizes preview, we might need another prop to resize code height too...

PR #1447, opened 2026-06-16
fix: preserve locale casing in content collection entry IDs

lirenjie95:
Closes #1446

The original issue was introduced in #1409

Issue #1446, opened 2026-06-16
[Tutorials] Tutorial content remains in English after switching language


luciawang0901:
### Most appropriate sections of the p5.js website?

Tutorials

### What is your operating system?

Windows

### Actual Behavior

When switching languages on the p5.js website's tutorial section, the page does not display the corresponding translated version. Instead, it still shows the English version.

### Expected Behavior

<img width="691" height="328" alt="Image" src="https://github.com/user-attachments/assets/d0477c83-b5d9-4ddd-8373-cb926d6b4a65" />

<img width="1848" height="866" alt="Image" src="https://github.com/user-attachments/assets/bcf41251-1d59-4e1f-a2b3-98c1a5929736" />

### Steps to reproduce

1. Go to https://p5js.org/tutorials/
2. Click on any tutorial, e.g., "Setting Up Your Environment"
3. Switch the language to "中文（简体）" using the language selector
4. Observe that the page content remains in English (or does not switch to the Chinese version)

### Would you like to work on the issue?

I'm afraid I can't work on this issue. I just wanted to report the bug. Please assign it to someone who can fix it. Thanks!

lirenjie95:
I discussed this issue with @luciawang0901 offline and I realized it would be a bug. Thanks her for pointing this out.

The problem started after the site upgraded to Astro 6. Astro 6 changed how content files get their IDs. It now lowercases the folder names, so zh-Hans becomes zh-hans. The website looks for the exact folder name zh-Hans to decide which language to show. Because the ID is now zh-hans, it can't find the Chinese translations and shows the English version instead.

How to fix it: tell Astro's glob() loader to keep the original file path instead of lowercasing it. For example:

```JavaScript
generateId: ({ entry }) => entry.replace(/\.(mdx|yaml)$/, "")
```

This should be added to the collection configs for tutorials, examples, reference, text-detail, and any other place using glob() with translated content.

PR #1474, opened 2026-06-30
fix: broken example URLs in search index(#1472)

### LINKED ISSUE #1472: Searching for `Rotate` brings examples that return 404 page

Other (specify if possible)

Mac OS

### Web browser and version

Firefox 152.0.3

### Actual Behavior

When searching for documentation for ’rotate’
https://p5js.org/search/?term=rotate

The search page returns results out of the examples pages and links to https://p5js.org/examples/transformation-rotate

That page returns a 404

### Expected Behavior

The expected result would be to point to the existing Rotate example page https://p5js.org/examples/Transformation-Rotate/

### Steps to reproduce

_No response_

Can be assigned to someone else

slash-init:
fixes #1472 

The search index builder was generating example URLs using its own slug transformation logic, including lowercasing the generated slug. This caused search results to point to URLs like `/examples/transformation-rotate`, while the actual routes are generated using `exampleContentSlugToLegacyWebsiteSlug`, resulting in broken links.
This PR removes the duplicated slug generation logic from the search builder and reuses `exampleContentSlugToLegacyWebsiteSlug` instead, so search indexing and example routing generate the same canonical URLs.
Since the search index builder is a standalone Node script and `_utils.ts` depends on Astro-specific modules, the helper was moved to `_utils-node.ts` and re-exported from `_utils.ts` to keep existing imports unchanged.


lirenjie95 (approved):
I've reviewed this PR and confirmed it resolves the issue. The root cause appears to be the recent Astro 6 upgrade, which also led to the i18n issue #1446 (fixed in my PR #1447). Let's keep an eye out for any other bugs related to this upgrade.

src/scripts/builders/search.ts
```diff
@@ -223,19 +223,19 @@ export const generateSearchIndex = async (
...
+        const exampleSlug = file
           .replace("src/content/examples/", "")
-          .replace(".mdx", "")
-          .toLowerCase()
-          // TODO: Separate Astro utils from the exampleContentSlugToLegacyWebsiteSlug
-          .replace(/^[\w-]+?\//, "")
-          .replace(/\d+_(.*?)\/\d+_(.*?)\/description$/, "$1-$2")
-          .replace(/_/g, "-");
-        relativeUrl = `/examples/${relativeUrl}`;
+          .replace(".mdx", "");
```
lirenjie95, line 233: We can add an assertion in test/scripts/build-search.test.ts for the examples case (e.g. expect /examples/Shapes-And-Color-Shape-Primitives) to guard against case/slug regressions.

ksen0: +1, this would be great,  @slash-init could you add please?

slash-init:
I've added the regression test addressing your suggestion. Please take a look.

lirenjie95 (approved):

ksen0:
Thanks both! I've also confirmed locally. I'll merge it here, on 2.0, and then update the search indeces manually in both. (Typically they are build during release.)

PR #1475, opened 2026-06-30
add p5.nanokontrol2 add-on library

SableRaf:
A p5.js add-on library for the Korg nanoKONTROL2 MIDI controller.

Repo: https://github.com/SableRaf/p5js_nanoKontrol2

Demo: https://openprocessing.org/@SableRaph/2973319


ksen0:
Thanks! If it works with 2.0 please do make a PR on `2.0` as well.

SableRaf:
Oh, actually it currently _only_ works with p5.js v2. Sorry I didn't realise I was on the v1 branch.

ksen0:
@SableRaf I noticed that among the hardware category it's the only one having to do with music. It's also in the listing not visible because there's alreayd multiple libraries in hardware. However there's only one add-on under sound, and I was intuitively expecting this to be under sound. Just a thought, please feel free to recategorize or ignore!

SableRaf:
I can see why you'd think that but p5.nanokontrol2 is for mapping sketch parameters to physical inputs on the Korg nanoKONTROL2 via webMIDI. It has nothing to do with sound/music :)

See https://www.instagram.com/reel/DaI5EeYt3XL/

PR #1490, opened 2026-07-01
fix: vertically center dropdown label

### LINKED ISSUE #1482: The "Language" and "Accessibility" dropdowns are not vertically centered

Home

Mac OS

### Web browser and version

Chrome

### Actual Behavior

The icon and the label are not vertically centered to the dropdown background:
Image

### Expected Behavior

It seems like the icons and the caret arrows are aligned vertically as expected, but the label is not. 

### Steps to reproduce

_No response_

No

lirenjie95:
Fixes #1482

## Problem
In the collapsed Language / Accessibility dropdowns, the icon and caret are `position: absolute` (manually offset to look centered), but the label is a normal in-flow `<span>` and `.selected` had no vertical centering — so the label sat too high relative to the icon and caret.

## Fix
Flex-center the button's in-flow content in `src/components/Dropdown/styles.module.scss`:

```scss
.selected {
  display: flex;
  align-items: center;
}
```

Because the icon and caret are absolutely positioned (out of flow), only the label is affected — the already-aligned icon and caret don't move.

## Testing
Verified in the browser that the label is now vertically centered with the icon and caret (tablet/desktop breakpoints and the expanded dropdown state).

## Screenshot
screenshot

### DIFF
--- a/src/components/Dropdown/styles.module.scss
+++ b/src/components/Dropdown/styles.module.scss
@@ -20,6 +20,8 @@
   width: 100%;
   height: 100%;
   cursor: pointer;
+  display: flex;
+  align-items: center;
   @media (hover: hover) and (pointer: fine) {
     &:hover {
       border: 1px solid var(--accent-color);
--- a/src/components/Dropdown/styles.module.scss
+++ b/src/components/Dropdown/styles.module.scss
@@ -62,6 +62,8 @@
 
   button {
     flex: 1;
+    display: flex;
+    align-items: center;
     text-align: left;
     background: none;
     border: none;
--- a/src/components/Dropdown/styles.module.scss
+++ b/src/components/Dropdown/styles.module.scss
@@ -22,8 +22,8 @@
   cursor: pointer;
   display: flex;
   align-items: center;
-  gap: 12px;
-  padding: 0 10px;
+  gap: var(--spacing-xs);
+  padding: 0 var(--spacing-xs);
   @media (hover: hover) and (pointer: fine) {
     &:hover {
       border: 1px solid var(--accent-color);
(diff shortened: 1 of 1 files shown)


doradocodes:
@lirenjie95 We tested the fix and the label looks vertically centered now! I think you could also remove the absolute positioning on the icons as well, and readjust the spacing. Thanks!

lirenjie95:
@doradocodes Done — the collapsed dropdown now uses a flex row instead of absolute positioning, so the icon, label and chevron align automatically. Kept the change minimal and left the expanded list as-is. Pushed in the latest commit.

doradocodes:
@lirenjie95 One more thing: I realized that all the items in the dropdown also have the same issue. Could you also apply the same fix to the items?
Screenshot 2026-07-03 at 11 52 52 PM

src/components/Dropdown/styles.module.scss
```diff
@@ -20,13 +20,26 @@
   width: 100%;
   height: 100%;
   cursor: pointer;
+  display: flex;
+  align-items: center;
+  gap: 12px;
+  padding: 0 10px;
```
doradocodes: nit: can we use some of the variables from https://github.com/processing/p5.js-website/blob/main/styles/variables.scss?

lirenjie95:
@doradocodes Good catch — applied the same flex layout to the dropdown items as well. The icon and label in each option are now vertically centered.

src/components/Dropdown/styles.module.scss
lirenjie95: Replaced all hardcoded spacing values introduced in this PR with variables from variables.scss (--spacing-xs and --spacing-sm).

lirenjie95:
Also ported this change to main in #1500

doradocodes:
@lirenjie95 I saw that the flexbox change was applied to the dropdown items as well, but I noticed that there was still a small misalignment that didn't fully allow the items to be vertically centered. Looking into it, I saw it was caused by the label being a `<button>` wrapped in a `<span>`. Locally, I applied `display: flex; align-items: center;` to the `.option button` selector in the `styles.module.css` and it is now fully vertically aligned. 

Here's a side-by-side comparison:
Screenshot 2026-07-10 at 12 01 04 PM

With the extra `display:flex;` applied:
Screenshot 2026-07-10 at 12 00 32 PM

So I think you can add that `display: flex; align-items: center;` to the `.option button` selector and this PR will be good to go!

lirenjie95:
Added for both #1490 and #1500. Thank you for the advice! @doradocodes

ksen0 (approved):

PR #1494, opened 2026-07-02
Optimized file checks via concurrent Promise.all and cleaned code.

tushar743-ui:
**Formatting:** Normalized inconsistent spacing .

**Parallelized existence checks:** Replaced the sequential for...of loop awaiting exists() one sketch at a time with Promise.all(allSketches.map(...)), so all file-existence checks run concurrently instead of blocking one-by-one.

**Simplified return:** Filtered allSketches directly against the resolved existence results instead of building availableSketches via push-in-a-loop, then dropped the pointless return [...availableSketches] spread in favor of returning the array directly.

**Why:** the original loop serialized an I/O-bound check (exists) per sketch, which is unnecessarily slow when the list grows; batching with Promise.all cuts wall-clock time while keeping identical output.

### DIFF
--- a/src/api/OpenProcessing.ts
+++ b/src/api/OpenProcessing.ts
@@ -53,7 +53,7 @@ export type OpenProcessingSketchResponse = {
 };
 
 // Selected Sketches from the 2025 curation
-export const priorityIds = ['2690038', '2484739', '2688829', '2689119', '2690571', '2690405','2684408' , '2693274', '2693345', '2691712']
+export const priorityIds = ['2690038', '2484739', '2688829', '2689119', '2690571', '2690405', '2684408', '2693274', '2693345', '2691712']
 
 async function exists(p: string) {
   try {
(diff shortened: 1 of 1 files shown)


Nwakaego-Ego:
Hi @tushar743-ui, thank you for this PR. I looked into it and wanted to share some thoughts.

Looking at the live site at `p5js.org/community/`, the sketches load without any noticeable delay for users. The `exists()` function imports from `node:fs/promises`, which confirms it runs at build time, not when a user visits the page.

Since no issue was opened for this, it is hard to know what specific problem is being solved. The p5.js website has a contributing process where changes like this should start with an issue. I may be missing context, so opening an issue would help the community weigh in on whether this optimisation is needed and the best approach to take.

Please check the [README](https://github.com/processing/p5.js-website/blob/main/README.md) for more on the contributing process.

cc  @doradocodes @ksen0

ksen0:
This is a build script, the serial loading is intentional to make it slightly easier to understand failures if the build doens't work; this build script also can fail for various reasons and already hs a draft PR as a refactor. As @Nwakaego-Ego described, there is no issue that exists for users here. For those 2 reasons I will close. Pleasde be sure to follow guidelines in the future, thank you!

PR #1491, opened 2026-07-01
Fix search bar placeholder font-size to match Language/Accessibility dropdowns.

AchieverSana:
Fixes #1481

The "Search" placeholder text was larger than the "Language" and "Accessibility" dropdown labels on desktop. Added the `text-base` Tailwind class to the search input to match the `font-size: 1rem` used by the dropdown component, so all three now render at the same size.

PR #1502, opened 2026-07-06
fix: mobile accessibility dropdown height (#1480)

### LINKED ISSUE #1480: [2.0] On mobile, the Language and Accessibility dropdowns have misaligned font

Home

Mac OS

### Web browser and version

Chrome

### Actual Behavior

1. Open the [beta.p5js.org](https://beta.p5js.org/) website on mobile view
2. See the "Language" and "Accessibility" dropdowns:
Image

### Expected Behavior

The "Language" and "Accessibility" dropdowns should have the equal font size as the Search bar. For the "Accessibility" dropdown, the text should not overlap with the background. 

Image

### Steps to reproduce

_No response_

No

slash-init:
## Summary

On mobile, the Accessibility dropdown was being placed into an implicit grid row, which defaulted to `auto` height. This caused the dropdown to appear compressed compared to the Language dropdown and Search bar.
This change adds `grid-auto-rows: var(--settings-container-height-touch)` so implicit grid rows use the same height as the explicitly defined rows, keeping all controls consistently sized on mobile.

Before :
image

After:
image

Fixes #1480


Nwakaego-Ego:
Hi @slash-init, thank you for this PR. I tested it locally, and here is what I found.

The height fix is correct and confirmed working. Adding `grid-auto-rows: var(--settings-container-height-touch)` ensures the Accessibility dropdown falls into an explicitly sized implicit grid row instead of defaulting to `auto` height. All three controls, Language, Accessibility, and Search, are now consistently sized on mobile. 

Issue still remaining

The font size mismatch mentioned in issue #1480 is still not addressed. After inspecting: 
- Search bar `font-size: 16px` inherited from parent (no explicit font size set)
- Language dropdown `font-size: 0.75rem` (12px) hardcoded
- Accessibility dropdown `font-size: 0.75rem` (12px) hardcoded

The dropdown text is hardcoded and set to 0.75rem (12px), which overrides inheritance and makes it smaller than the Search bar. The fix would be to either remove the hardcoded `font-size: .75rem` so the dropdowns inherit from the parent like Search does, or set it explicitly to match.  @doradocodes, your thoughts on this?

slash-init:
Hey @Nwakaego-Ego, Thanks for the review!
I think the font size mismatch is already being addressed in PR #1491

doradocodes:
@slash-init You're right, and that PR has been merged now. I will merge your change now as well.

doradocodes:
@slash-init Actually, upon further testing, the font-sizes are mismatched on mobile. This is what Ego was mentioned in her comment above.

Mobile view on the p5js.org:
Screenshot 2026-07-10 at 12 20 42 PM

And the expected result should be like this (which is how it looks on v1.p5js.org):
Screenshot 2026-07-10 at 12 20 35 PM

I opened another Issue for this: https://github.com/processing/p5.js-website/issues/1507. If you'd like to continue working on it, leave me a comment on the new Issue and I will assign it to you!

PR #1503, opened 2026-07-07
Adding Ancient Brain to Education Resources page

ehcsav:
Ancient Brain is a free platform for writing and running code in the browser. I recently ported Kevin Workman's Happy Coding course onto it with permission. Other p5.js docs and tutorials are already on the site, including a port of Daniel Shiffman's Coding Train. Learners can clone, modify, and share what they make directly on Ancient Brain.


Nwakaego-Ego:
cc @ksen0

ksen0 (approved):

doradocodes:
@ksen0 In this PR, Ancient Brain was added at the beginning of the list under "English education resources from the community". Is this ok, or should it be added at the end of this section? Doesn't look like there's any alphabetically sorting for this section either.

ksen0:
@doradocodes I think it's ok for new resources to be added generally at the top. Maybe we can add a note suggesting something like that? Ideally we would have a better organization overall but "newest first" seems reasonbale and more or less what's been done recently...

Issue #1025, opened 2025-11-10
"Smoke Particles" example emits many FES messages


ksen0:
### Most appropriate sections of the p5.js website?

Examples

### What is your operating system?

Mac OS

### Actual Behavior

The [smoke particles (beta/2.0)](https://beta.p5js.org/examples/math-and-physics-smoke-particle-system/) example, when pasted into the p5.js Editor with any 2.x version, runs smoothly.

### Expected Behavior

The example will cause too many FES errors to be runnable

### Steps to reproduce

Here is the sketch: https://editor.p5js.org/ksen0/sketches/Z-DcIih0s

### Would you like to work on the issue?

Sure!

ksen0:
@davepagurek @perminder-17 I'm actually not sure whether this is a usage issue or a bug issue, please feel free to transfer to the p5.js repository if incorrectly filed! Regression on 2.x

davepagurek:
It looks like the sketch is using tint(color, alpha). It looks like in the docs for both 1.x and 2.x, that signature isn't supported, just tint(color). I would need to check if we actually support that usage or not, to see whether we just need to update the docs to add that, or if we need to correct the example. but we probably shouldn't be continuously logging every frame like that either, maybe checking if we've already logged an identical FES message and not logging it again if so?

VANSH3104:
@ksen0, can I try this? Could you please assign it to me?

ksen0:
Hey @VANSH3104 sorry I missed this but I know you're busy with svg work, so I'll open it to another volunteer, I hope that's fine!

I'll reserve it for the CodeDay microinternship. For CodeDay participants, you can:
* make a PR against `dev-2.0` branch of this repo (or `main` if filing after July 1: please [check status here](https://github.com/processing/p5.js-website/issues/1426)) updating the example code to use correct call if needed
* optionally: make a PR against `p5.js` repo `main` branch to execute on Dave's idea above to avoid continuous FES logging. More about Friendly Error System (FES) [here](https://p5js.org/contribute/friendly_error_system/).

kaissy-a:
I'm working on this.

ksen0:
@kaissy-a can you confirm please if you're with a CodeDay team? thanks!

kaissy-a:
Hi, yes, they assigned my team to this issue.

kaissy-a:
[...]

PR #1522, opened 2026-07-19
Fix: Smoke particle example

kaissy-a:
Issue: [1025](https://github.com/processing/p5.js-website/issues/1025)

Changes: Added parameters to createVector.

Changes running here: https://editor.p5js.org/kaissy/sketches/Xovljq5vg 

Team Members:
@kaissy-a
@esoloriovilla-dot
@RyRy241


limzykenneth:
Why is `length` renamed `len` here? I don't think there would be anything that would conflict with the name `length`?

kaissy-a:
Hi! We noticed during testing in the p5.js web editor that the original variable name threw a redeclaration error (Function "length" on line 32 is being redeclared and conflicts with a p5.js function). With the user in mind we thought renaming the variable would help avoid conflict with the global length property. Please let us know if you prefer a different naming convention or approach!

ksen0:
Thanks for flagging that @kaissy-a, it is indeed a v2 behavior that was not present in v1. It's not intended so I will look into whether it should be reported as a bug, because there is no global length function in the pubic API, and multiple examples use `length` as a var name, which might cause other weirder errors/behaviors in other examples or in the future.

@davepagurek @limzykenneth this seems to be a strands shadow problem introduced in v2, should I make a bug issue in p5.js repo?

image

@kaissy-a let's rename it back to `length` in this example though because this shouldn't be happening. Really helpful you spotted this weird behavior, thank you

ksen0:
@kaissy-a thanks; this is still marked as "draft" - intentional or ready from your side?

kaissy-a:
Hi, sorry for the late response. Yes this is intentional I wanted to wait on my team to see if there is anything else to add before marking it ready for review, since it's currently only 1 line of code.

PR #1525, opened 2026-07-20
Restore full hero image rotation for non-English locales

### LINKED ISSUE #1524: Hero images on p5js.org localized pages

Home

None

### Web browser and version

_No response_

### Actual Behavior

Going to https://p5js.org/ko/ will show one hero image related to the most recent p5 version. However, going to https://p5js.org/ will randomly show different photographs.

### Expected Behavior

The random photographs should be available in all languages. Please use the `v1` branch to find translations for captions where available

### Steps to reproduce

_No response_

someone else can take it!

NisargJasani0602:
Resolves #1524

The homepage hero images rotate randomly, but that only happens when there's more than one image in the list for that locale. en.yaml has all 9 community photos, but es, hi, ko, and zh-Hans got cut down to just one image a while back when the beta site was pointing at a specific version release, and nobody added the rest back after that. So those locales have just been stuck on one static image since.

This puts the same 9 photos back in all four locale files, and keeps the current version image in there too instead of removing it.

Captions for these photos have never been translated into other languages, even going back through older versions of the site, so I just left them in English like they already were.

### DIFF
--- a/src/content/homepage/es.yaml
+++ b/src/content/homepage/es.yaml
@@ -2,8 +2,44 @@ title: p5.js
 heroText: >
   p5.js es una herramienta amigable para aprender a programar y hacer arte. Es una biblioteca de JavaScript libre y de código abierto construida por una comunidad inclusiva y solidaria. ¡p5.js da la bienvenida a artistas, diseñadores, principiantes, educadores y cualquier otra persona!
 heroImages:
+  - image: ./images/hero.webp
+    linkTarget: "https://archive.org/details/processing-community-catalog-2021"
+    altText: Lauren Lee McCarthy reading the Processing Community Catalog.
+    caption: "Lauren Lee McCarthy reading the Processing Community Catalog. Photo credit: Maximo Xtravaganza."
+  - image: ./images/p5for50plus.webp
+    linkTarget: ""
+    altText: Inhwa Yeom teaching p5.js to people aged 50+.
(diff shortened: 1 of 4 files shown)


Nwakaego-Ego:
Hi @NisargJasani0602, thank you for this PR. I verified the following before testing:

Checked v1 branch: all 4 locale files (es, hi, ko, zh-Hans) already had the full 9 heroImages, and captions were never translated. So keeping captions in English is the right call.

Checked main branch: confirmed the bug. All 4 locale files had only 1 heroImage, while en.yaml had all 9. No rotation was possible for non-English users.

After testing locally, hero image rotation now works correctly on all non-English locale pages. 

cc  @doradocodes @ksen0

ksen0:
Thanks @Nwakaego-Ego  for review! @NisargJasani0602 Looks good, please remove the "2.3" version hero image, this actually was not intended to remain in any ofthe languages (also not in englihs). After that will be good to go! Thanks both.

NisargJasani0602:
Hi @ksen0 I have removed the 2.3 version hero image mentions from the yaml file. If it needs any more changes, let me know. Also the folder still contains the image: src/content/homepage/images/2.3.png. I have not removed it since a 2.2 version image also exists. Thanks!

doradocodes:
Looks good to me, will merge this. @ksen0 Do you think that we should create a new issue to translate the alt text of these images?

ksen0:
@doradocodes great idea yes please feel free!

Issue #1505, opened 2026-07-08
[A11y] Inline links not visually distinct from text


ritiupa:
axe-core flags link-in-text-block (serious) on /reference/ and /reference/p5/circle/. Inline links inside body text are only distinguishable by color. There is no underline or other visual indicator. This fails WCAG 1.4.1 (Use of Color).

Examples:
The p5.sound link in the intro paragraph on /reference/
The ellipseMode() link inside body text on /reference/p5/circle/

Steps to reproduce:
1. Go to /reference/ or /reference/p5/circle/
2. Run axe DevTools or npm run test:a11y
3. See link-in-text-block violation

Expected: inline links should be visually distinguishable without relying on color alone, for example by adding an underline.

I searched open and closed issues before filing. No existing issue covers this rule in this repo.

@coseeian, flagging for label review.

coseeian:
Hi @ritiupa, thanks for bringing this up.

I’m actually aware of this one - it’s been on my radar. The reason I haven't opened a dedicated issue for it yet is that, while adding an underline is trivial from a purely engineering standpoint, the broader scope requires a discussion with the p5.js product owner. We need to evaluate whether making underlines the default is acceptable to the design direction of the entire site, or if we can potentially introduce an accessibility preference toggle, similar to what GitHub does.

Because of these larger product decision scope, I wanted to avoid opening a public issue that contributors might immediately jump into fixing before we have a defined direction.

I’ve assigned this to myself for now. I'll find time to work on this and sync with the product owners about their thoughts.

ksen0:
Thanks @coseeian and @ritiupa , I've added this to the Usability&Design issue. To my knowledge, nothing is in the way, we could add underlines. However, it would be good to get some feedback on this. I will share this question with the PF team and on the p5js-web and maintainers discord channels. If there are no objections before end of day (anywhere on earth) July 20th, then I think this work can be unblocked. If there are any objections then we can consider how to address them in a way that ultimately improvs site accessibility.

I think if there are other problems like this - simple to execute and relatively isolated but pending a decision - we can approach in the same way: Make a "discussion" issue that does not accept volunteers before a decision is reached; share widely for a week; proceed when there is consensus. Though it would be good not to open too many such discussion issues simultaneously, if possible.

limzykenneth:
Links currently have underline on when hovered over, is the idea to switch that to always be underlined instead?

ksen0:
[...]

PR #1528, opened 2026-07-22
feat: add underline to links within text

coseeian:
To address: https://github.com/processing/p5.js-website/issues/1505 .

### Summary

- `styles/markdown.scss`: add persistent `text-decoration: underline` to prose links inside `.rendered-markdown`, scoped as `a:not(.button-link)` so styled button links are excluded.

### DIFF
--- a/src/components/EducationResource/index.astro
+++ b/src/components/EducationResource/index.astro
@@ -25,7 +25,9 @@ const authorHTML = (await Astro.slots.render("author")).replace(/\<p[^\>]*\>/gm,
       <div set:html={authorHTML} />
       <slot name="description" />
     </div>
-    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline [&_a]:outline-1 [&_a]:outline-[var(--type-magenta-dark)]">
+    <!-- Materials render as outlined pills, so opt them out of the underline
+         that .rendered-markdown links get -->
+    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline [&_a]:outline-1 [&_a]:outline-[var(--type-magenta-dark)] [&_a]:!no-underline">
       <slot name="materials" />
     </div>
   </div>
--- a/src/components/EducationResource/index.astro
+++ b/src/components/EducationResource/index.astro
@@ -25,9 +25,11 @@ const authorHTML = (await Astro.slots.render("author")).replace(/\<p[^\>]*\>/gm,
       <div set:html={authorHTML} />
       <slot name="description" />
     </div>
-    <!-- Materials render as outlined pills, so opt them out of the underline
-         that .rendered-markdown links get -->
-    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline [&_a]:outline-1 [&_a]:outline-[var(--type-magenta-dark)] [&_a]:!no-underline">
+    <!-- Materials render as outlined pills, so drop the persistent underline
+         that .rendered-markdown links get, but keep the hover underline. The
+         !important is needed to beat .rendered-markdown, and would otherwise
+         kill :hover too -- hence the second utility. -->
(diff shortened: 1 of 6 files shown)


limzykenneth:
They are mostly there but there are some false negatives:

These probably should be underlined as well.
Screenshot 2026-07-22 at 12 00 28

These in community sketches too perhaps
Screenshot 2026-07-22 at 12 03 11

Should these be underlined too?
Screenshot 2026-07-22 at 12 04 04

I wonder if there is somewhere in the CSS rules that applies the p5 magenta highlight we can add the underline rule at the same place then we'll catch more or less the places we want to have underlines? Unless that introduce too many false positives.

coseeian:
Let me take a look.

coseeian:
Hi,

I found there wasn't one existing place to put it. The magenta link was being applied in 4 different ways across pages. 

So I created `.text-link` in `styles/global.scss`. The single place now has one rule where the color and the underline are declared together. 

`.text-link` works on a link itself or on any container (styling its descendant links). Every call site now points at that rule instead of redeclaring the color, and `markdown.scss` no longer duplicates the style.

Let me know if any concern about this approach, thanks. :)

coseeian:
I verified both the Reference and Sketches pages locally, across light and dark modes - links now include underlines:

Screenshot 2026-07-27 at 18 54 54

Screenshot 2026-07-27 at 18 41 54

For the About page, axe-core only flags links inside body text, but I updated it as well for design consistency. Let me know if you prefer keeping it as-is!

Screenshot 2026-07-27 at 18 42 54

src/layouts/SketchLayout.astro
```diff
@@ -114,13 +114,9 @@ const iframeTitle = `OpenProcessing Sketch: ${title} by ${authorName}`;
...
     {instructions && <p class="text-md my-sm md:my-lg">{instructions}</p>}
 
-    <p class="text-xs md:text-base mb-3xl">
-      This <a class="text-type-magenta" href={makeSketchLinkUrl(sketchIdNumber)}
-        >sketch</a
-      > is ported from the <a
-        class="text-type-magenta"
-        href="https://openprocessing.org">OpenProcessing</a
-      >
+    <p class="text-xs md:text-base mb-3xl text-link">
```
coseeian, line 117: Note: This updates the color to a slightly darker shade of magenta in light mode. I made this change to better align with the design across the rest of the site, but please let me know if you’d prefer to stick with the original shade. Thanks!

src/components/EducationResource/index.astro
```diff
@@ -25,7 +25,11 @@ const authorHTML = (await Astro.slots.render("author")).replace(/\<p[^\>]*\>/gm,
       <div set:html={authorHTML} />
       <slot name="description" />
     </div>
-    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline [&_a]:outline-1 [&_a]:outline-[var(--type-magenta-dark)]">
+    <!-- Materials render as outlined pills, so drop the persistent underline
+         that .rendered-markdown links get, but keep the hover underline. The
+         !important is needed to beat .rendered-markdown, and would otherwise
+         kill :hover too -- hence the second utility. -->
+    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline [&_a]:outline-1 [&_a]:outline-[var(--type-magenta-dark)] [&_a]:!no-underline [&_a]:hover:!underline">
```
coseeian: Note: Added `[&_a]:!no-underline [&_a]:hover:!underline` to maintain current production behavior.

https://github.com/user-attachments/assets/c09733af-2223-47dc-b1cf-9be506fb0199

Applied directly here since it's a one-off, but happy to extract it into a shared utility like `.button-link` if preferred.

limzykenneth:
```suggestion
    <div class="text-sm leading-8 [&_a]:whitespace-nowrap [&_a]:rounded-full [&_a]:mr-1 [&_a]:py-1 [&_a]:px-2 [&_a]:outline-1 [&_a]:outline-(--type-magenta-dark) [&_a]:no-underline! [&_a]:hover:underline!">
```

Just fixing some lint warnings

limzykenneth (commented):
Just one inline. The rest looks good.

src/components/EducationResource/index.astro
coseeian: Looks good, updated: https://github.com/processing/p5.js-website/pull/1528/commits/7595f80f62bfc4bdff02af766a95b86bdb82af6d . 

Thanks! 👍

src/layouts/SketchLayout.astro
ksen0: That seems fine to me!

src/components/EducationResource/index.astro
ksen0: Great catch, thank you!

ksen0 (approved):
Thanks for this! OK to merge I think

Issue #1242, opened 2026-03-16
Website search should be improved


ksen0:
### Increasing Access

Currently, the keyword search on the website often does not help users find what they look for. This is evident in usability testing, and in issue reports. To make the information on the site more accessible, search should be improved.

### Most appropriate sub-area of p5.js?

Other (specify if possible)

### Feature request details

- Here is the current search script: https://github.com/processing/p5.js-website/blob/main/src/scripts/builders/search.ts
- There is also a filter on reference: https://github.com/processing/p5.js-website/blob/main/src/components/ReferenceDirectoryWithFilter/index.tsx
- Commit introducing main portion of current search: https://github.com/processing/p5.js-website/commit/993e6c59a4eda4d5f0819e5b0a9fdd85e039fedc
- This issue is for research for what's possible technically: comments and draft PRs that demonstrate an idea are welcome from different people! But it is not yet ready for focused work. But keep in mind if you make a PR it may be closed.

### Related issues

* https://github.com/processing/p5.js-website/issues/493 search through descriptions
* https://github.com/processing/p5.js-website/issues/459 search field suggestions dropdown feature
* https://github.com/processing/p5.js-website/issues/1249 searchProvider bug

ksen0:
This was discussed in the March 16, 2026 call in the Discord/#p5js-website working group (including @davepagurek @kunstewi @skyash-dev @aashu2006 @nbogie @doradocodes @Nwakaego-Ego). Below are some of the topics:

## Categories are not searcheable/filtreable in Reference

Category titles are not included in the index, because they're not pages (maybe they should be pages, or maybe they should be included in index somehow else) so [searching for "dom" will not get p5.dom](http://p5js.org/search/?term=dom).

A couple weeks ago in #help there was also a discussion of the Reference "filter" function behaving in the same unintuitive way:

In this example the filter function in the reference was used, same behavior as search on "dom"

<img width="300" alt="screenshot of p5.dom filter term returning nothing in the reference" src="https://github.com/user-attachments/assets/3b00879d-8d20-44e6-877c-c881cc1b2493" /> 

<img width="300" alt="screenshot of dom filter term returning unrelated terms like random" src="https://github.com/user-attachments/assets/00f58867-cfde-42e9-93ba-e58ef2680013" />

[...]

Issue #1221, opened 2026-03-09
Usability, Design & Implementation Research - start here!


ksen0:
# Start here for:

* Ideas on design or user experience improvements, including usability bugs
* Research and discussion for implementing larger design changes or ideas

### Topic

During the first few months of 2026, two groups of students from [a graduate course on usability studies](https://www.smunson.com/teaching/hcde517/w25/) are working on usability research for the p5.js website. This analysis includes user studies. The result will inform what work will be prioritized on the website, and opened to contribution.

Updates will be shared on this issue, so you can "watch" it if interested. Many thanks to all who have already participated in the user studies, and everyone who is interested but has not yet shared their experience will be invited to discuss the outcomes of the research when they are shared.

If you're a designer, educator, and/or anyone who uses the website and would like to be part of this effort, I hope you stay tuned here!

Another way to get involved and help right away is to look through past open issues or PRs, especially ones that got stuck in review and/or are proposing significant changes to design or function of the site (for example: related to the "search" function), and mention them in the comments below. 

### Research Reports

(more coming soon, both reports and further detail, analysis, and data)

[1] [Presentation by Abdulrahman Alobaid, Bill White, Chad Brandon, Su Min Kim](https://docs.google.com/presentation/d/1dKfShyyCxpPTGhiLR5JfAsX0IqK_gxmZlbdR-MwuB5w/edit?slide=id.p#slide=id.p)
[2] TBA - 2nd report
[3] TBA - detailed analysis

Besides analyzing the site as it is, I hope these reports can help us test usability in the future. Additional to reports the students have also shared their protocols and other materials, which we can reuse, and which I will link.

Please reference the reports in the comments using "[1]" or "[2]". I'll keep the numbers above consistent. Thank you!

ksen0:
@SableRaf's comments:

1. (over Discord) on [1] recommendation of educational resource being made top-level: "I’d suggest making it a “Teach” page rather than “educational resources” to keep it consistent with the rest of the menu."
2. Consider adding Fathom analytics

My additional observations:

1. Another note: the reference breadcrumb is not meaningful for p5.sound.
2. Offline download should be reinstated

Nishthajain7:
[...]

PR #1531, opened 2026-07-24
[do not merge] Add privacy-first tracking

ksen0:
Relates to / partially addresses https://github.com/processing/p5.js-website/issues/1242 and https://github.com/processing/p5.js-website/issues/1221 - part of working on this issue, Fathom Analytics has been setup. Fathom Analytics is a privacy-first tool that [does not use cookies](https://usefathom.com/docs/troubleshooting/cookies). It therefore has a few limitations relative to something like google analytics, but I believe it can address all or most of our actual needs. We do not currenlty use google analytics on the reference site (though it is used on the p5.js editor)

I am approaching these trackers like in a video game / game analytics. Running and editable code sketch block anywhere on the site is a conversion event on a per-user basis, so we can see % of unique users who did that.

Same for spending a certain amount of time focused on a page, or achieving a relative scroll depth. It is possible to use a secondary variable to see relative scroll depth or focus time conversion per user on each page:

image

Finally, specific to the Search topic, terms are logged only when there are exact matches (string appear in the reference). Otherwise, they are redacted as "[FUZZY]", because that _might include personal information_. While unlikely, this follows the principle of **data minimization**: unless we have a good reason to collect the data, might as well not. Right now, there is no data. This would already be much better.

In all cases, the number of search results is included as a second component of the string, so the events can be exported and analyzed. 

image

## Possible other changes

Maybe scroll % should only trigger after 30s? If someone scrolls up and down in the first 30s does that "count?" We should also see if we can notice such a pattern in the aggregate data.

Possibly search term count buckets should be applied, like with the others, such as:

* Search <10 [term]
* Search <50 [term]
* Search <100 [term]
* Search >100 [term]

Other Focus and Scroll buckets can be used, especially if informed by the previous studies about whats "enough" time. Scroll % would be useful on other parts of the site, too, but in my opinion it is better to roll these analytics out based on what actually corresponds to an active question.

### DIFF
--- a/src/components/ReferenceTracker/index.tsx
+++ b/src/components/ReferenceTracker/index.tsx
@@ -1,0 +1,74 @@
+import { useEffect, useRef } from "preact/hooks";
+
+const TIME_LABELS: Record<number, string> = {
+  10000: "Focus 10s",
+  30000: "Focus 30s",
+  90000: "Focus 90s",
+  300000: "Focus 5min",
+  600000: "Focus 10min",
+};
+
+const SCROLL_LABELS: Record<number, string> = {
+  25: "Scroll 25%",
+  50: "Scroll 50%",
+  75: "Scroll 75%",
+  100: "Scroll 100%",
+};
+
+const track = (label: string) => {
+  // Fathom analytics does not use cookies. The data collected
+  // is aggregate and minimal. The scope is currently only on
+  // the Reference page, but it can be generalized in the future.
+  if (window.fathom && window.location.pathname.startsWith("/reference/")) {
+    window.fathom.trackEvent(label);
+  }
+};
+
+const ReferenceTracker = ({
+  timeThresholds,
+  scrollThresholds,
+}: {
+  timeThresholds: number[];
+  scrollThresholds: number[];
+}) => {
+  const hasFiredTime = useRef<Set<number>>(new Set());
+  const hasFiredScroll = useRef<Set<number>>(new Set());
+  const elapsed = useRef(0);
+
+  useEffect(() => {
+    const tick = () => {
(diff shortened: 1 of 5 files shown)


ksen0:
https://github.com/processing/p5.js-website/issues/1401 has been merged, tracking should also be added to filter. I'll have to think a bit more about it, but maybe tracking terms that get results and only when theres "enough" of a pause during typing; in this case I don't think it's needed to track number of results

ksen0:
Feedback is welcome! But please **do not merge** we are checking if we need to update privacy policy to make this usage clear.

src/components/ReferenceTracker/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  50: "Scroll 50%",
+  75: "Scroll 75%",
+  100: "Scroll 100%",
+};
+
+const track = (label: string) => {
+  // Fathom analytics does not use cookies. The data collected
+  // is aggregate and minimal. The scope is currently only on
+  // the Reference page, but it can be generalized in the future.
+  if (window.fathom && window.location.pathname.startsWith("/reference/")) {
```
lirenjie95, line 22: This pathname check misses all localized reference pages. Non-default locales are served under `/[locale]/reference/`...(e.g. `/es/reference/p5/line`), so `pathname.startsWith("/reference/")` is false there and none of these events will ever fire on non-English reference pages — which likely skews the data toward English-speaking users. Consider stripping the locale prefix before testing, e.g. with the existing `removeLocalePrefix` util from `@i18n/utils`.

ksen0: Thanks for the catch!!

PR #1539, opened 2026-07-27
fix: include descriptions for reference entries in search index

### LINKED ISSUE #1537: Reference entries in the generated search index omit their `description` field

Reference

Linux

### Web browser and version

Firefox v153.0

### Actual Behavior

Reference pages that contain a `description` in their source content are generated without a `description` field in the search index.
While building the search index, the `reference` case checks that `data.description` exists:

```ts
if (!data.description) {
  continue;
}
```
However, unlike other content types, it never assigns that value to the local `description` variable before the search entry is created.
Later, the search entry is generated as:

```ts
searchIndex[title] = {
  relativeUrl,
  description,
  ...(alias ? { alias } : {}),
};
```
Since `description` was never assigned for reference entries, its value remains `undefined`. During JSON serialization, properties with the value `undefined` are omitted, so the generated search index does not contain a `description` field for reference pages.

### Expected Behavior

Reference entries should include their `description` in the generated search index, just like other supported content types.
For example, a reference page such as `mouseX` already contains a description in its source content, and that description should be included in the generated search index.

### Steps to reproduce

1. Open `src/scripts/builders/search.ts`.
2. Inspect the `case "reference"` block.
3. Observe that it checks `data.description` but never assigns it to the local `description` variable.
4. Build the search index.
5. Inspect the generated search index JSON.
6. Compare a reference entry (for example, `mouseX`) with other content types such as libraries or contributor docs.
7. Notice that the reference entry does not contain a `description` field even though the source file defines one.

Yes

slash-init:
This PR fixes an issue where reference entries in the generated search index were missing their `description` field. While the builder verified that a description existed, it never assigned it before creating the search index entry, causing the field to be omitted from the generated JSON and preventing reference descriptions from being indexed for search.

fixes #1537

### DIFF
--- a/src/scripts/builders/search.ts
+++ b/src/scripts/builders/search.ts
@@ -256,6 +256,7 @@ export const generateSearchIndex = async (
         if (!data.description) {
           continue;
         }
+        description = data.description;
         break;
       case "libraries":
         title = data.name;


doradocodes:
@slash-init This looks good! Merging now.

slash-init:
@doradocodes Thanks! Just a doubt, don't we have to add a regression test for this bug? I asked the same in the issue thread.

doradocodes:
@slash-init answered your question in the original thread! Thanks for double-checking

PR #1548, opened 2026-07-31
test: add regression test for reference descriptions

slash-init:
This PR adds a regression test for the recently fixed issue where reference entries in the generated search index omitted their `description` field.
The test verifies that reference entries include their expected description in the generated search index, helping prevent similar regressions in the future.

Follow-up to #1539.


doradocodes:
Looks good, merging this now. Thanks for your work!

@Nwakaego-Ego can you create a branch to merge this into `v1` as well?

PR #1519, opened 2026-07-17
test(styles): regression test for markdown max-width (#1452)

JoseRodriguez26:
## Summary

Adds a regression test guarding against the tutorial-layout bug reported in #1452, where the GLSL tutorial (`/tutorials/intro-to-glsl/`) — and other rendered-markdown pages — collapsed into a thin, unreadable vertical column of text.

**The bug itself is already fixed** (by #1457). This PR does **not** re-fix it; it adds the missing guardrail so the bug — and its whole class — cannot silently return, and documents the root cause so the "it broke then healed itself" mystery is explained.

Co-authored with @julian-men.

## Root cause & timeline (answers "why did it break ~2 weeks ago but not today?")

The width cap on `.rendered-markdown > *` in [`styles/markdown.scss`](styles/markdown.scss) is what keeps tutorial text at a readable column width.

| State | The line | Works? | Effect |
|-------|----------|--------|--------|
| Before | `@apply max-w-screen-md;` | ✅ | Text capped, renders normally |
| Broken | `@reference max-w-screen-md;` | ❌ | No cap → vertical column |
| Fixed (now) | `max-width: variables.$breakpoint-tablet;` | ✅ | Text capped at 770px |

- **2026-05-27 — broke:** commit `adb24d3e4` "Upgrade Astro to v6" (a large upgrade, almost certainly an automated codemod) rewrote `@apply max-w-screen-md;` → `@reference max-w-screen-md;`. In Tailwind v4, `@reference` is only valid pointing at a *file* (`@reference "app.css";`); pointed at a *utility name* it is invalid and silently dropped by the browser — no error. The max-width cap vanished and the layout collapsed. **This is #1452.**
- **2026-05-27 → 2026-06-21 — broken window (~3.5 weeks):** matches the maintainer's "broken a couple weeks ago, fine now" report.
- **2026-06-21 — fixed:** commit `3ed344be6` "Fix tutorial content too wide" (#1457) replaced the broken line with `max-width: variables.$breakpoint-tablet;`. That PR targeted a *different* symptom (content too wide) and healed #1452 as a side effect — which is why the fix was never connected to this issue.

## What this PR adds

`test/styles/markdown-scss.test.ts` (Vitest, 5 tests). It compiles `styles/markdown.scss` with `sass` and asserts:

1. Rendered content has a real, positive `max-width` (not missing/zero).
2. That max-width equals `$breakpoint-tablet`.
3. `.full-width` correctly opts out to a wider limit.
[...]

Issue #1452, opened 2026-06-17
Shader tutorial continues to have vertical text problem on beta


ksen0:
### Most appropriate sections of the p5.js website?

Home

### What is your operating system?

None

### Actual Behavior

Scroll in this tutorial on the beta site (`2.0` branch) and observe the thin unreadable column of text: https://beta.p5js.org/tutorials/intro-to-glsl/

### Expected Behavior

The same tutorial on the 1.x side (`main` branch) does not have this problem: https://p5js.org/tutorials/intro-to-shaders/

### Steps to reproduce

Please note that the branch names will be updated soon, more info here: https://github.com/processing/p5.js-website/issues/1426

### Would you like to work on the issue?

someone else can!

ksen0:
This issue is reserved for CodeDay microinternship. Please check https://github.com/processing/p5.js-website/tree/main/docs to get started on the website, particularly tehcnical overview and content overview.

JoseRodriguez26:
Hi! I'm Jose Rodriguez, working through the Mentor/CodeDay summer program with my team. We'd like to claim this issue and start on it ASAP. Could it be assigned to me? Thanks!

ksen0:
It's yours @JoseRodriguez26 ! I just checked and the bug appears to be resolved - **however, a mysteries here is also not good!** Could you please add some strong tests to prevent weird regressions? Possibly also going back in the recent commits and figuring out why `2.0` branch had a problem 2 weeks ago but not today? Thanks in advance.

Please watch out for an upcoming branch rename and deployment update: https://github.com/processing/p5.js-website/issues/1426

julian-men:
Hi @ksen0,

We've (@JoseRodriguez26 and I) been looking into this issue together and wanted to share what we've found so far to make sure we're heading in the right direction.

So far, we were able to reproduce the bug on commit 5ab6db3, where the Intro to GLSL tutorial still displays the text in a thin vertical column. We then used git bisect and found that the issue is fixed in commit 3ed344b (PR #1457).

We also traced the regression back to the Astro v6 upgrade (adb24d3e), where the markdown width rule in styles/markdown.scss changed. The fix in 3ed344b replaces that with an explicit max-width rule.

Our current plan is to leave the stylesheet as-is since it's already fixed and focus on adding a regression test. We also want to verify that the test fails on the broken commit and passes on the current version.

Does this sound like the right direction, or would you recommend a different approach for testing this regression?

Thanks!

[...]

PR #1549, opened 2026-08-01
test(styles): add regression test for markdown max-width (#1452)

Nwakaego-Ego:
v1 cherry-pick of #1519.


Nwakaego-Ego:
cc @doradocodes

PR #1558, opened 2026-08-04
Add Spanish translation for Introduction to GLSL tutorial

xdroberto:
Adds the Spanish translation of the [Introduction to GLSL tutorial](https://p5js.org/tutorials/intro-to-glsl/), addressing the Spanish part of #1412 (the issue mentions `intro-to-shaders.mdx`, which was since renamed to `intro-to-glsl.mdx`). As discussed in that issue, I claimed the Spanish translation for this tutorial.

This creates `src/content/tutorials/es/` — it would be the first tutorial available in Spanish.

Notes on the translation:

- 1:1 structural parity with the English original (862 lines): all code blocks, components, and markup are untouched, except code comments and `describe()` strings, which are translated following the convention of the existing Korean tutorial translation.
- GLSL keywords (`attribute`, `varying`, `uniform`, etc.) are kept literal when referring to code, and rendered as Spanish prose terms elsewhere.
- Only `title`, `description`, and `featuredImageAlt` are translated in the frontmatter; the `{/* Note for contributors */}` comment stays in English.
- Validated locally with the dev server: `/es/tutorials/intro-to-glsl/` renders correctly and the tutorial appears listed on `/es/tutorials/`.

PR #1559, opened 2026-08-05
Add Spanish translation for Conditionals and Interactivity tutorial

xdroberto:
Adds the Spanish translation of the [Conditionals and Interactivity tutorial](https://p5js.org/tutorials/conditionals-and-interactivity/), addressing the Spanish part of #1379. This follows #1558 (Spanish translation of Introduction to GLSL).

Notes on the translation:

- 1:1 structural parity with the English original (915 lines): all code blocks, components, tables, and markup are untouched. Code comments and the placeholder text inside `AnnotatedLine` are translated, following the convention of the existing Korean and Chinese translations; code identifiers (including the `condition` placeholder in the syntax samples) are kept in English.
- Only `title`, `description`, and `featuredImageAlt` are translated in the frontmatter.
- The English source uses non-breaking spaces for indentation in several code blocks; these are normalized to regular spaces in the translation, as the Hindi translation of `variables-and-change.mdx` already does. This keeps the samples safe to copy into an editor.
- Validated locally with the dev server: `/es/tutorials/conditionals-and-interactivity/` renders correctly.

### Three errors in the English source

While translating, I found three places where the English text contradicts its own code samples. I wrote the technically correct version in Spanish rather than propagating the errors to Spanish readers, so I want to flag them explicitly:

1. **Line 700** — "Make the sun move by changing the `sunHeight` variable as long as it is **less than** 130", but the code right below is `if (sunHeight > 130)`. It should read *greater than* 130.
2. **Line 775** — "The value in `sunHeight` stops changing once the Boolean expression ... is `false`. This happens when `sunHeight` is any value **greater than** 130." It is the opposite: the expression becomes `false` when `sunHeight` is less than or equal to 130.
3. **Lines 846 and 848** — the logical-operators section asks for `sunHeight` "less than 130 *AND* greater than 480", and the sample reads `if (sunHeight < 130 && sunHeight > 480)`. That condition can never be true; based on the preceding nested conditional it should be `if (sunHeight > 130 && sunHeight < 480)`.

This last one is the only line of code in the file that differs from the English original. I am happy to open a separate PR fixing the English source (and to revert the Spanish to a literal translation instead) — whichever you prefer.

### DIFF
--- a/src/content/tutorials/es/conditionals-and-interactivity.mdx
+++ b/src/content/tutorials/es/conditionals-and-interactivity.mdx
@@ -1,0 +1,915 @@
+---
+title: "Condicionales e interactividad"
+description: Un tutorial sobre cómo usar sentencias condicionales y crear sketches interactivos.
+category: introduction
+categoryIndex: 3
+featuredImage: ../images/featured/ConditionalsA.png
+featuredImageAlt: Una vista previa de la animación de amanecer creada en el tutorial de Condicionales e interactividad.
+relatedContent:
+  examples:
+    - en/11_3d/06_framebuffer_blur/description
+  references:
+    - en/p5/if
+    - en/p5/mouseispressed
+    - en/p5/keycode
+    - en/p5/circle
+authors:
+  - Greg Benedis-Grab
+  - Layla Quiñones
+---
+
+import Callout from "../../../components/Callout/index.astro";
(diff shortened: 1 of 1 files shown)


Nwakaego-Ego:
Hi @xdroberto, thank you for this PR.

@marioguzzzman, please take a look when you get a chance. Thank you.

PR #1560, opened 2026-08-05
Add p5.sound CDN link to download page

AmanUllah687:
Part of processing/p5.js#9037

p5.sound is being removed from the p5.js complete library zip since it's
now maintained as a separate library. This adds a section to the download
page linking to the current p5.sound CDN build (using the existing but
previously-unused `cdnSoundUrl` global) and the jsDelivr package page.

### DIFF
--- a/src/content/text-detail/en/download.mdx
+++ b/src/content/text-detail/en/download.mdx
@@ -22,8 +22,7 @@ This link redirects you to the p5.js Editor online so you can begin using p5.js 
 <LinkButton variant='link' url={`https://editor.p5js.org?version=${p5Version}`}> p5.js Editor </LinkButton>
 
 ### Download the Complete Library
-This is a download containing the p5.js library file and an example project. It does not contain an editor. It also does not contain p5.sound.js; please see below instructions for sound support.
-
+This is a download containing the p5.js library file and an example project. It does not contain an editor. If this download includes a p5.sound.js file, please disregard it — it is an outdated version incompatible with p5.js 2.x. See below for up-to-date sound support instructions.
 <LinkButton variant='download' url={fullDownloadUrl}> Complete Library </LinkButton>
 
 ### Download Single Files
--- a/src/content/text-detail/en/download.mdx
+++ b/src/content/text-detail/en/download.mdx
@@ -5,6 +5,7 @@ import LinkButton from '../../../components/LinkButton/index.astro'
 import CodeContainerWithCopy from "../../../components/CodeContainer/CodeContainerWithCopy.astro";
 import {
   cdnLibraryUrl,
+  cdnSoundUrl,
   fullDownloadUrl,
   libraryDownloadUrl,
   minifiedLibraryDownloadUrl,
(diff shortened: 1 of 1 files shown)


src/content/text-detail/en/download.mdx
```diff
@@ -21,7 +22,7 @@ This link redirects you to the p5.js Editor online so you can begin using p5.js
 <LinkButton variant='link' url={`https://editor.p5js.org?version=${p5Version}`}> p5.js Editor </LinkButton>
 
 ### Download the Complete Library
-This is a download containing the p5.js library file, the p5.sound addon, and an example project. It does not contain an editor. Visit [Get Started](/tutorials/get-started) to learn how to setup a p5.js project.
+This is a download containing the p5.js library file and an example project. It does not contain an editor. Visit [Get Started](/tutorials/get-started) to learn how to setup a p5.js project.
```
ksen0: How about: "It does not contain an editor. It also does not contain p5.sound.js; please see below instructions for sound support."

src/content/text-detail/en/download.mdx
```diff
@@ -33,6 +34,11 @@ These are downloads or links to the p5.js library file. No additional contents a
   <LinkButton variant='download' url={minifiedLibraryDownloadUrl}> p5.min.js </LinkButton>
 </div>
 
+### Sound Support with p5.sound
```
ksen0: The jsdelivr package is "p5.sound" but the library and repo name is `p5.sound.js`. Please use "p5.sound.js" to refer to it in the title and the text. Sorry for the inconvenience!

ksen0 (commented):
Thanks! Just 2 minor things then it will be ready to merge. @doradocodes please feel free t omerge if you get to it before I do and the comments are addressed, thank you in advance

src/content/text-detail/en/download.mdx
AmanUllah687: Updated, thanks! Now reads: "It does not contain an editor. It also does not contain p5.sound.js; please see below instructions for sound support."

src/content/text-detail/en/download.mdx
AmanUllah687: Updated, section is now titled "Sound Support with p5.sound.js" and the body text uses p5.sound.js consistently. Left the jsDelivr URL and link text pointing at the actual npm package name (p5.sound) since that's what jsDelivr calls it.

Nwakaego-Ego:
Hi @AmanUllah687, thank you for this PR. One little thing I noticed while reading through it: the download page now says the zip "does not contain p5.sound.js." From your comment on issue #9037 though, it sounds like the old p5.sound.js file is still bundled into the zip for now, and will only be removed once the follow-up PR in the p5.js repo lands. So the wording is a little ahead of itself,  describing something as already done when it's actually still one step away from being done.

Could you tweak the wording slightly so it matches where things stand right now? Thanks again for all the work on this.

AmanUllah687:
> Hi @AmanUllah687, thank you for this PR. One little thing I noticed while reading through it: the download page now says the zip "does not contain p5.sound.js." From your comment on issue #9037 though, it sounds like the old p5.sound.js file is still bundled into the zip for now, and will only be removed once the follow-up PR in the p5.js repo lands. So the wording is a little ahead of itself, describing something as already done when it's actually still one step away from being done.
> 
> Could you tweak the wording slightly so it matches where things stand right now? Thanks again for all the work on this.

Good catch, thank you, you're right, the old file is still in the zip until the p5.js repo PR lands. Updated the wording to reflect the current state accurately rather than getting ahead of things.

doradocodes:
@Nwakaego-Ego Please make a branch to merge into `v1`!

PR #1565, opened 2026-08-07
feat: setup local a11y playwright test for homepage

coseeian:
Hi everyone,

This PR picks back up on the automated accessibility testing work originally discussed in https://github.com/processing/p5.js-website/issues/827.

Earlier in https://github.com/processing/p5.js-website/pull/933, we put together initial playwright and axe-core settings and tests, but it looks like those changes didn't make it onto the current main branch. This PR re-introduces that foundation with just enough lightweight configuration to get the tests running smoothly on local environments again.

### What's Included
* Restored Dependencies: Re-adds playwright and axe-core baseline setups.
* Core Tests: Restores the initial batch of accessibility test suites (on EN homepage).
* Local Workflow: Re-establishes local test execution support so anyone can run quick local checks.

### Next Steps (Following PRs)
* Triage and fix any new accessibility issues identified by the automated scans.
* Will work on integrating automated scans directly into PR CI pipeline to catch potential regressions early.
* Expand test coverage across more pages and localized versions.

Feedback and suggestions are welcome! Thanks!

PR #1566, opened 2026-08-09
fix: render promo banner as complementary landmark, not contentinfo

coseeian:
This PR addresses a duplicate `contentinfo` landmark issue introduced in https://github.com/processing/p5.js-website/pull/929.

Because both the promo banner and the main site footer were rendered as root-level `<footer>` elements, two `contentinfo` landmarks were generated in the accessibility tree. This was flagged during [Playwright axe-core scan](https://github.com/processing/p5.js-website/pull/1565) ([axe rule details](https://dequeuniversity.com/rules/axe/4.12/landmark-no-duplicate-contentinfo?application=playwright)). While HTML permits multiple footers, ARIA best practices recommend maintaining a single `contentinfo` landmark per document so screen reader users aren't confused with multiple primary footers (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/contentinfo_role#one_contentinfo_landmark_per_page)).

Since the banner is a dismissible promo rather than main footer content, I think it should be reasonable to change the  landmark to `complementary` (using an HTML `<aside>`), which still keeps it easily accessible without conflicting with the main footer.

Visual styling remains unchanged. Feedback on the role or label choice is very welcome. Thanks!

### DIFF
--- a/src/components/Banner/index.astro
+++ b/src/components/Banner/index.astro
@@ -3,15 +3,15 @@ import { render } from 'astro:content';
 import { Icon } from "@components/Icon"
 const { entry } = Astro.props;
 const { Content } = await render(entry);
-const { link, title } = entry.data;
+const { link, title, label } = entry.data;
 ---
 
-<footer class="banner bg-accent-color text-accent-type-color" style={{ display: 'none' }} data-title={title} aria-label="sticky">
+<aside class="banner bg-accent-color text-accent-type-color" style={{ display: 'none' }} data-title={title} aria-label={label}>
   <a class="banner-content" href={link} target="_blank">
     <Content />
   </a>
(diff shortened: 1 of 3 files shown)


ksen0:
Nice, this makes a lot of sense

ksen0 (approved):

coseeian:
Hi @ksen0,

Quick nudge on this PR. No rush, but whenever you have a free moment to merge it, that would be awesome - I’m getting ready to open [the next PR about playwright scan on CI](https://github.com/processing/p5.js-website/pull/1567) building on top of this work.

Thanks again for the review!

ksen0:
Thanks for the bump, will merge

PR #1572, opened 2026-08-12
Fix framebuffer blur example for p5.js 2.x

### LINKED ISSUE #1571: [p5.js 2.0+ Bug Report]: Framebuffer example breaks on 2+

- [ ] Accessibility
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [x] WebGL
- [ ] WebGPU
- [ ] p5.strands
- [ ] Build process
- [ ] Unit testing
- [ ] Internationalization
- [ ] Friendly errors
- [ ] Other (specify if possible)

### p5.js version

2+

### Web browser and version

chrome

### Operating system

macos

### Steps:
1. https://beta.p5js.org/examples/3d-framebuffer-blur/ observe somethign strange happening (intended: https://p5js.org/examples/3d-framebuffer-blur/). Instead of row of rotating spheres there's a smear of crescent shapes that fades into nothing
2. if pasting code into p5.js editor, and select 2.3 in version, a whole other also strange behavior (also no syntax highlighting after 2.2?)

BHARATH0153:
## Summary

Fixes the 3D Framebuffer Blur example, which breaks on p5.js 2.x (see processing/p5.js#8940). The example was transferred here because the fix lives in the example itself, not the library.

## What changed

- **Scoped the blur shader.** Wrapped the final `shader(blur); plane(...)` pass in `push()`/`pop()` so `userFillShader` no longer leaks into the next frame's lit sphere pass (which caused the crescent smear / fading feedback loop in 2.x).
- **Smaller spheres.** Shrunk the spheres to `min(width/8, 35)` so the row reads clearly at the smaller sizes the website renders.
- **Arc in depth.** Spheres now sit at slightly different distances from the camera (z offset up to 70), so nearest/farthest spheres visibly fall out of focus, matching the example description.
- **Shallower depth of field.** Retuned the blur shader's focal point (`0.9` to `0.82`) and blur slope (`40` to `150`) so the DOF effect is visible given the compressed depth range of the default camera.

closes #1571

### DIFF
--- a/src/content/examples/en/11_3D/06_Framebuffer_Blur/code.js
+++ b/src/content/examples/en/11_3D/06_Framebuffer_Blur/code.js
@@ -25,9 +25,10 @@ precision highp float;
 varying vec2 vTexCoord;
 uniform sampler2D img;
 uniform sampler2D depth;
-float getBlurriness(float d, float focalDepth) {
-  // Blur more the farther away we go from the focal point
-  return abs(d - focalDepth) * 200.;
+float getBlurriness(float d) {
+  // Blur more the farther away we go from the
+  // focal point at depth=0.9
+  return abs(d - 0.9) * 80.;
 }
 float maxBlurDistance(float blurriness) {
   // Cap the blur radius so far-away pixels don't smear
@@ -37,11 +38,8 @@ float maxBlurDistance(float blurriness) {
 void main() {
   vec4 color = texture2D(img, vTexCoord);
   float samples = 1.;
-  // Use the depth at the center of the canvas as the focal point,
-  // where the middle sphere sits, so it stays in focus
-  float focalDepth = texture2D(depth, vec2(0.5, 0.5)).r;
   float centerDepth = texture2D(depth, vTexCoord).r;
-  float blurriness = getBlurriness(centerDepth, focalDepth);
+  float blurriness = getBlurriness(centerDepth);
   for (int sample = 0; sample < 20; sample++) {
     // Sample nearby pixels in a spiral going out from the
     // current pixel
@@ -55,7 +53,7 @@ void main() {
 
     // How far should its blur reach?
     float sampleBlurDistance =
-      maxBlurDistance(getBlurriness(sampleDepth, focalDepth));
+      maxBlurDistance(getBlurriness(sampleDepth));
 
     // If it's in front of the current pixel, or its blur overlaps
     // with the current pixel, add its color to the average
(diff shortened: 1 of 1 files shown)


davepagurek:
Do you have a screenshot or screen recording of what this looks like in context now?

ksen0:
Hi @BHARATH0153 I've come across your recent comments and this PR, and it does seem like there are some unverified LLM outputs in comments and code. Because there was also a past temporary ban, if there are any other future violations of contributor guidelines, community norms, or AI usage policy, I may have to apply a permanent ban. Thank you for your time, and understanding.

BHARATH0153:
@davepagurek 
screenshot

davepagurek:
Is that the size it previews at on the site? it appears much smaller on the current site. Also as it stands, I don't think it currently achieves the same effect (of a focal blur) because it's spreading out the samples so much. That amount could be ok at a smaller window size, which is why it's important to test on the website repo locally. Regardless it could be a good idea to add a max offset size to prevent this.

BHARATH0153:
blur_1000x700
blur_360x360
blur_640x480

davepagurek:
I'm not sure how much the z offset adds, so it may not be worth complicating the logic with it.

Testing your code in the example (you can edit the code on the live site and rerun it) it seems like nothing is in focus. Ideally the center of the line of spheres is in focus so it shows how the depth is involved in the filter:


https://github.com/user-attachments/assets/ccf6d3a3-3aa7-4ff2-a65f-f7ce31db9019

BHARATH0153:
https://github.com/user-attachments/assets/73787f2c-c00e-4a37-8ff9-e3fdc1dcddb0

BHARATH0153:
rwec

davepagurek:
I think grabbing the z value from the center of the screen will cause the focus to shift suddenly as the line of spheres rotates in front. 

It also looks like the z updates weren't removed. It's hard to tell from the screen recording what the updates look like in context, and your screenshot of the actual size on the website doesn't show anything of the sketch, so on its own it's not adding anything to the conversation. We aren't asking for screenshots just for the sake of it, but to help communicate and reach a shared goal faster. We expect you to think of it through that lens, asking yourself what someone else will get from the information you're providing, and making sure you're giving them something they can understand and work with.

ksen0:
Hi @BHARATH0153 your activity around this issue and PR have consistently created additional workload for reviewers. We expect contributors to sense-check their work before asking for feedback. In another case, we would temporarily suspend an account, sin e this kind of activity creates delays for everyone. However, in this case, following guidelines and AI policy has been a repeated problem. Therefore, a permanent suspension will follow. Thank you for understanding

PR #1574, opened 2026-08-14
Add Fathom click tracking for reference filter vs global search usage

ritiupa:
## Summary

Adds two Fathom custom click events so we can directly compare reference filter usage against sitewide search usage in the analytics dashboard, building on the base Fathom script added in #1511.

* `Reference Filter Used`: fires once per session on the first non-empty keystroke in the page-level "Filter by keyword" box on /reference/.
* `Global Search Used`: fires once per session on the first non-empty sitewide search term.

Both events are guarded behind `typeof window !== "undefined" && window.fathom`, so nothing breaks if the Fathom script hasn't loaded yet or is blocked.

## Why

Related to #1221 (Fathom analytics was one of the requested usability/design improvements). Right now filter and search usage aren't distinguishable in analytics, which makes it hard to tell how often each tool is actually used, something this round of usability research keeps running into.

## Testing

* 14/14 existing tests passing
* Full type check run, no other areas of the site affected
* Manually verified both events fire exactly once per session and don't fire on empty input

## Notes

I shared this branch with @ksen0 for a look before opening the PR (no changes requested), opening it now so it's in the queue for review.

### DIFF
--- a/src/components/ReferenceDirectoryWithFilter/index.tsx
+++ b/src/components/ReferenceDirectoryWithFilter/index.tsx
@@ -186,6 +186,7 @@ export const ReferenceDirectoryWithFilter = ({
 }: ReferenceDirectoryWithFilterProps) => {
   const [searchKeyword, setSearchKeyword] = useState("");
   const inputRef = useRef<HTMLInputElement>(null);
+  const hasTrackedFilterUse = useRef(false);
 
   const filteredEntries = useMemo(
     () => filterCategoryData(categoryData, searchKeyword),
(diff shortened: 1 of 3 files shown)


ksen0:
Hi @ritiupa , I apologize for the miscommunication, but please do check the existing PR about this first: https://github.com/processing/p5.js-website/pull/1531 Please do have a look and leave a review there.

After looking into it, we will only merge behavioral tracking once privy policy is updated.

Since this one is a duplicate, unfortunately I will close it! Sorry!

PR #1576, opened 2026-08-14
Add p5.are.na to the libraries list

jake-welch-design:
Adds p5.are.na, which allows blocks from Are.na channels to be loaded directly into p5.js sketches. Compatible with p5 1.x via preload() and 2.x via async setup(). MIT license, on npm, includes eight runnable examples each with its own p5 web editor sketch. Image is 1500x1000.


ksen0:
Very cool, thank you! Since it also supports 1.x, please feel free to make a matching PR against `v1` branch which serves the v1 archive reference

Issue #1445, opened 2026-06-16
On deploy, update an OpenProcessing curation


ksen0:
### Increasing Access

Currently, the examples on the p5.js web editor [must be manually updated](https://github.com/processing/p5.js-web-editor/issues/4106), and this makes new examples not accessible/maintainable.

### Most appropriate sub-area of p5.js?

Examples

### Feature request details

On deployment of the main branch, using env vars to specify curation and auth, update an OpenProcesing curation via [the API](https://www.postman.com/openprocessing-api/openprocessing-api/collection/am4ysl7/openprocessing-public-api?sideView=agentMode) to include all the [examples](https://beta.p5js.org/examples/)

Note that the website should still be the source of truth for examples, because the p5js website has multiple languages and OP currently does not.

ksen0:
This issue is currently reserved for CodeDay, please do not make PRs. Thanks for your understanding!

Yisakor-Mirany:
I am working on this issue, and I am from CodeDay Labs.

Yisakor-Mirany:
Hi! I'm working on this issue for CodeDay. I inspected the OpenProcessing Public API documentation and found the read endpoints for curations (GET /api/curation/{id} and GET /api/curation/{id}/sketches), but I couldn't find any documented write endpoint for updating a curation or adding sketches. Is there a private/internal endpoint or another API that this issue is intended to use?

ksen0:
Hi @Yisakor-Mirany ! Sorry for the delay in answering and the missing info. Although there is no public documentation, the endpoint does exist (and will be publicly documented soon). With thanks to OP developer Sinan who shared the info about it:

```
/**
     * Add a sketch to a curation.
     * Endpoint: POST /api/curation/{id}/sketches/{visualID}
     * (Routed via routes.php to api/curation_sketch/$1/$2.)
     * Mirrors Curation_controller::submit_ajax authorization, dedup, status and
     * notification behavior.
     *
     * @apiPublic
     */
```

Here is an example test curation I made on my own account and added a sketch to programmatically: https://openprocessing.org/curation/91135 I used `POST https://openprocessing.org/api/curation/91135/sketches/2934070` with no payload and with Authentication: Bearer [token] header using my account token. Same URL with `DELETE` deletes the sketch.

For working on this, please create a text curation where only owner can submit, and put your token into a secret on your fork (never share your tokens!). Curation id should also be a repo-level variable that is different for your fork, rather than hardcoded.

[...]

PR #1577, opened 2026-08-16
Add OpenProcessing example synchronization tooling

Yisakor-Mirany:
## Summary

This PR begins implementing #1445 by adding tooling to synchronize p5.js website examples with an OpenProcessing curation.

The current implementation loads the English examples from the p5.js website, prepares their code and attribution data, connects to the OpenProcessing API, and compares the website examples against a configured OpenProcessing curation.

## What changed

* Added an OpenProcessing API client.
* Added a script for comparing p5.js examples with an OpenProcessing curation.
* Added reusable example attribution utilities.
* Added attribution comments to prepared sketch code.
* Added support for `OPENPROCESSING_TOKEN` and `OPENPROCESSING_CURATION_ID` environment variables.
* Added `npm run sync:openprocessing`.
* Added tests for OpenProcessing configuration and example attribution.
* Kept API credentials outside of the source code.

## Current behavior

The sync command currently performs a safe dry run:

```bash
npm run sync:openprocessing
```

It:

1. Loads the English p5.js examples.
2. Reads their metadata and sketch code.
3. Adds attribution information to the prepared sketch code.
4. Fetches the configured OpenProcessing curation.
5. Reports examples missing from the curation and sketches that no longer correspond to website examples.

No OpenProcessing sketches are modified during the dry run.

## Testing

Full test suite:

```text
Test Files  13 passed (13)
Tests       64 passed (64)
```

Project checks:

```text
0 errors
0 warnings
```

I also tested the sync script against a separate OpenProcessing test curation.

### OpenProcessing sync dry run

Screenshot 2026-08-13 at 20 10 36

### Automated tests

Screenshot 2026-08-13 at 20 11 21

## Configuration

The integration expects the following environment variables:

```text
OPENPROCESSING_TOKEN
OPENPROCESSING_CURATION_ID
```

No API credentials are committed to the repository.

## Remaining work / feedback requested

The current PR establishes the synchronization, comparison, API, and attribution foundation.

I would especially appreciate guidance on the intended OpenProcessing API endpoint/payload for creating or updating the actual sketches before adding them to the curation by `visualID`.

Once that flow is confirmed, the synchronization can be extended to create/update missing sketches and remove outdated entries from the curation.

Related to #1445

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "build:search": "tsx ./src/scripts/builders/search.ts",
     "build:p5-version": "tsx ./src/scripts/p5-version.ts",
     "custom:dev": "tsx ./src/scripts/branchTest.ts",
-    "custom:cleanup": "tsx ./src/scripts/resetBranchTest.ts"
+    "custom:cleanup": "tsx ./src/scripts/resetBranchTest.ts",
+    "sync:openprocessing": "tsx ./src/scripts/sync-openprocessing.ts"
   },
   "engines": {
     "node": ">=22.0.0"
--- a/src/layouts/ExampleLayout.astro
+++ b/src/layouts/ExampleLayout.astro
@@ -1,17 +1,23 @@
 ---
 import type { CollectionEntry } from "astro:content";
 import { render } from "astro:content";
+
+import EditableSketch from "@components/EditableSketch/index.astro";
 import Head from "@components/Head/index.astro";
+import RelatedItems from "@components/RelatedItems/index.astro";
+
 import { getCurrentLocale, getUiTranslator } from "../i18n/utils";
 import {
   generateJumpToState,
+  getFallbackRemixData,
   getRelatedEntriesinCollection,
 } from "../pages/_utils";
-import BaseLayout from "./BaseLayout.astro";
-import EditableSketch from "@components/EditableSketch/index.astro";
-import RelatedItems from "@components/RelatedItems/index.astro";
+import {
+  EXAMPLE_ATTRIBUTION,
+  getExampleAttributionData,
+} from "../utils/exampleAttribution";
 
-import { getFallbackRemixData } from "../pages/_utils";
+import BaseLayout from "./BaseLayout.astro";
 
 interface Props {
   example: CollectionEntry<"examples">;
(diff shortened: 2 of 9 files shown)


Nwakaego-Ego:
Hi @Yisakor-Mirany  Thank you for your work so far on this PR. While testing "npm run sync:openprocessing" on Windows, I encountered the following error: ENOENT: no such file or directory, open ...\p5js-website-2.0\code.js
The paths returned from "findDescriptionFiles()" contain Windows "\" separators, while "lastIndexOf("/")" searches for "/". This returns "-1" and causes "codePath" to point to "code.js" at the repository root. Would you be open to using "dirname(descriptionPath)" with "join()" here so Node can handle the path separator for each operating system? It may also be helpful to add a test covering a Windows-style path. Let me know what you think.

I also checked the attribution requirement from #1445. The new exampleAttribution.ts utility and the existing src/layouts/ExampleLayout.astro file each work out the authors, collective attribution year, and licence separately. This means the same attribution rules are still written in two places.

Since #1445 asks for these rules to be defined once, could they be moved into one shared function that is used by both the website layout and the OpenProcessing sketch comments? Thank you.

Yisakor-Mirany:
Hi @Nwakaego-Ego . Thank you for the detailed feedback! I’ve addressed both points. I updated the path handling to use dirname() and join() so it works correctly across operating systems and added test coverage for Windows-style paths. I also refactored the attribution logic so the shared rules are defined in one place and reused by both the website layout and OpenProcessing synchronization.

I pushed the updates to the PR, and all tests/checks are now passing. Please let me know if there’s anything else you’d like me to adjust.

Nwakaego-Ego:
Thank you, @Yisakor-Mirany, for your work on this so far. npm run sync:openprocessing now finds all 65 English examples before stopping because OPENPROCESSING_TOKEN is not set. This confirms that the original Windows issue that caused the script to look for code.js at the repository root has been fixed. There is one related Windows path issue remaining. The example ID is still created with:

const id = descriptionPath
  .replace(`${EXAMPLES_DIRECTORY}/`, "")
  .replace("/description.mdx", "");

On Windows, descriptionPath contains \, so these replacements do not match. I confirmed locally that the ID remains the full path: src\content\examples\en\07_Repetition\03_Kaleidoscope\description.mdx instead of: 07_Repetition/03_Kaleidoscope

The current dry run does not use this ID. If it is intended for the full synchronization, it should be the same on every operating system. Could you create it with path tools that work across operating systems, such as dirname() and relative(), make sure the final ID uses the same format everywhere, and add a Windows test for it?

PR #123, opened 2024-04-03
Update contribute docs and script

outofambit:
closes #121 

Fixes a couple issues in the build contribute script:
- moved script file to fit directory organization of other scripts
- cleans out previous generated docs when run
- handles markdown to MDX comment conversion reliably

This PR also pulls in the latest updates from the p5 repo's contribute docs

Issue #1404, opened 2026-05-19
[GSoC 2026]: Continued Development of Translation Tracker


aashishpanthi:
### Topic

Hi everyone,
I'm Aashish, this year's GSoC Contributor, working on the continued development of the translation tracker. I wanted to share and potentially get feedback on what I'll be working on this summer.

> I’ll be updating these items as the project progresses. Feedback is always welcome!

### Description
This tracking issue outlines the goals, deliverables, and timeline for the Google Summer of Code 2026 project: **Continued Development of Translation Tracker**. 

Building upon the robust foundation laid during GSoC 2025, this project aims to close operational gaps in our automated translation synchronization workflow. The core objective is to make tracking highly actionable for human translators, automate issue management, expand our content coverage, and provide clear progress visualization for language stewards and maintainers.

---

#### 1. Improve Existing Issue Quality
- [x] **In-Issue Diffs:** Update the tracker (`index.js`) to parse the `englishDiff` patch snippet and render it within a collapsible `<details>` section inside the GitHub Issue body. This allows volunteer translators to see exactly what changed without navigating away.
- [ ] **Duplicate Detection:** Leverage the GitHub REST API to search for open tracker issues matching the specific file path and language labels before creating a new tracking ticket, updating the existing issue with new diff info if found.

#### 2. Stub-File Generation for New Pages
- [x] **Automated Placeholders:** When a new English MDX page is created without a language counterpart, the GitHub Action will automatically generate a placeholder stub file in the respective language folders.
- [x] **Frontmatter & Tracking Flags:** Stubs will pre-fill essential frontmatter (e.g., `title`, `description`) in English and append an explicit tracking flag (e.g., `needsTranslation: true`).
- [x] **Pull Request Workflow:** Group these stub creations into a single automated PR per language. This avoids auto-merging and leaves full review control in the hands of repository maintainers.

#### 3. Auto-Close Issues on Translation PR Merge
- [x] **Automated Cleanup:** Build a dedicated GitHub Action workflow triggered on `pull_request` (closed + merged) events to prevent manual backlog noise.
- [x] **Path & Label Parsing:** The workflow will parse changed translation file paths, identify target languages, search for corresponding open tracker issues (via labels like `lang-es`, `needs translation`), and programmatically close them. It will natively support standard "Resolves #<issue>" conventions.

#### 4. Translation Progress Visualization
[...]

PR #1473, opened 2026-06-29
Add automated translation stub generation for missing content

aashishpanthi:
Addresses #1404

### Changes

This PR implements **stub-file generation** for the GSoC 2026 Translation Tracker ([#1404](https://github.com/processing/p5.js-website/issues/1404)). When a new English content files (except references) exists without translations, the tracker generates placeholder MDX stubs and opens **one PR per language** for maintainer review.

#### Features

**Stub generation**
- Detects missing translations for files inside content (`findMissingTranslations`)
- Generates stub MDX files from English sources:
  - Copies essential frontmatter (`title`, `module`, `submodule`, `file`, `description`)
  - Sets `needsTranslation: true`
  - Adds a placeholder body and HTML comment linking to the English source
- Opens **one PR per language** via the GitHub API (blob → tree → commit → branch → pull request)
- Default languages: `es`, `hi`, `ko`, `zh-Hans`
- `STUB_MAX_FILES` applies **per language** (default `50`), not as a global cap across all languages

**GitHub Actions**
- New workflow: `.github/workflows/translation-stubs.yml`
  - Runs in stub mode (`GENERATE_STUBS=true`), separate from issue tracking in `translation-sync.yml`
  - Triggered on push when `src/content/**/en/**` changes (except `src/content/references/en/**` for now)
  - Supports manual `workflow_dispatch` with `full_scan` and `languages` inputs

**Local testing**
- `npm run test:stubs` — dry-run stub generation into `stub-preview/` (no PRs, no `src/content/` changes)

**Code organization**
After discussing with Divyansh, I refactored the translation tracker from a single large `index.js` into focused modules:

| File | Re-purposed to |
|------|----------------|
| `index.js` | Entry point and orchestration |
| `constants.js` | Supported languages, content types, stub frontmatter keys |
| `utils.js` | Path helpers, frontmatter parsing, file scanning |
| `github-tracker.js` | GitHub API client (issues, diffs, stub PR creation) |
| `workflows.js` | Translation status checks and stub generation logic |

#### Design decisions

- Stub mode and issue-tracking mode are **mutually exclusive** (`GENERATE_STUBS=true` returns before issue logic), so PR and issue workflows stay independent.
- Stub PRs are **never auto-merged**; language stewards and maintainers review them.
- Except reference content only for now; `examples`, `tutorials`, etc. have been added via `STUB_CONTENT_TYPES`.

### Test plan

- [x] Verified on fork (`aap009/p5.js-website`): workflow creates stub PRs when English content files are added
[...]

PR #1582, opened 2026-08-23
Translation tracker: auto close issue on PR merged and closed

aashishpanthi:
## Summary

The translation tracker opens a GitHub issue for every outdated or missing translation. Until now, those issues stayed open after the translation actually landed, so maintainers had to close them by hand and there was backlog.

This adds a `translation-auto-close.yml` workflow that runs when a PR touching translated content is merged. It works out which tracker issues the PR resolves, marks off the languages that were translated, and closes each issue once nothing is left to translate.

Addresses #1404.

## How an issue is found

Relying on `Resolves #123` did not worked well with real contributors: most translation PRs either use no linking keyword or mention the issue somewhere other than the body. So any `#123` is treated as a candidate, and it is looked for in:

- the PR title and body
- review summaries (maintainers can/do write "this handles `#123`" during review)
- issues that cross-reference the PR from their own thread
- pasted `https://github.com/<owner>/<repo>/issues/123` links

> Also, comments written by bots are ignored, so the tracker's own automation cannot feed itself.

## How false positives are avoided

Matching every `#123` is deliberately greedy, so candidates are filtered:

- fenced code blocks and inline code spans are stripped first, which removes most hex colors and code samples
- `owner/repo#123` cross-repo references and `#123abc` are not matched; numbers are capped at five digits because color codes are six digits and we are in four digits, we'll reach six digits in years **(needs attention)**
- the PR's own number is skipped
- each remaining candidate must exist, must not be a pull request, and must carry the `needs translation` label

## How an issue is updated

Which languages a PR translated is derived from its changed file paths, for example `src/content/examples/es/...` means Spanish. For each such language, the action removes the `lang-<code>` label and strikes that language's line in the issue body. The issue is closed only when no `lang-*` labels are left.

One tracker issue often covers several languages for the same English file, which creates a wrinkle: GitHub closes an issue natively the instant a PR saying `Fixes #123` is merged, even when only one of four languages was done. The action detects this and reopens the issue, leaving a comment naming the languages still working. 

## Testing

`npm run test:auto-close` runs the flow against a mocked GitHub API and covers: keyword-free references, the reopen path, comment-sourced references, bot filtering, cross-references, noise rejection, referenced pull requests, and PRs with no translation files.


aashishpanthi:
Note: This branches off the stub-file-generation PR. So, #1473 should be merged first, then this.

.github/workflows/translation-auto-close.yml
```diff
@@ -0,0 +1,41 @@
+name: Translation Auto-Close
+
+on:
+  pull_request:
```
Divyansh013, line 4: Hi @aashishpanthi , I see one blocker here for which I'm not very sure. 

For any PR opened from a fork (which is how virtually all external translation contributors submit), GitHub downgrades `GITHUB_TOKEN` to read-only regardless of the permissions: block. This workflow needs issues: write to remove labels, edit the body, comment, and close/reopen, all of those API calls (removeLabel, issues.update, createComment) will return 403 for fork PRs.

The whole motivation of the PR ("real contributors" not using linking keywords) implies fork-based PRs, so as written the feature might work only for maintainers pushing branches inside the repo.


Can we test this on a fork? 

cc: @ksen0  do you have any views here?

.github/actions/translation-tracker/auto-close.js
```diff
@@ -0,0 +1,310 @@
+const core = require('@actions/core');
+const { GitHubCommitTracker } = require('./github-tracker');
+const { collectReferencedIssues } = require('./pr-references');
+const { SUPPORTED_LANGUAGES } = require('./constants');
+const { getLanguageDisplayName, getTranslationPath } = require('./utils');
+
+// List all files changed in a merged PR (paginated).
+async function listPullRequestFiles(octokit, owner, repo, pullNumber) {
```
Divyansh013, line 8: This function returns every file including status: "removed"/"renamed", and neither `identifyLanguagesFromFiles` nor the `verifiedLanguages` filter checks status:


const verifiedLanguages = languages.filter((lang) => {
  const paths = expectedPaths.get(lang);
  if (!paths || paths.length === 0) return false;
  return paths.some((p) => changedFileNames.includes(p)); // matches removed files too
});

A PR that deletes src/content/examples/es/foo.mdx would strike Spanish, remove lang-es, and potentially close the tracker issue - the opposite of the truth. Filter to file.status === 'added' || 'modified' (or exclude removed) before deriving languages and verifying paths.

Divyansh013 (changes requested):

PR #1591, opened 2026-09-02
Add back p5.embroider library

nkymut:
Hi @ksen0 

Updated p5.embroidery to support p5.js v2.x and set it as the default. 
It should now be ready to be listed on the website again.

Also created a migration guide of p5.embroider for v1.x users.
https://github.com/nkymut/p5.embroider/blob/main/MIGRATION.md


ksen0:
Thank you so much, for sharing the updated library and also for that thorough migration guide!

PR #1595, opened 2026-09-07
Updates from 2.3.3 patch

ksen0:
The contributor docs build script (` npm run build:contributor-docs`) is changing image paths, not sure when this was introduced, but please do not merge, this version breaks.


src/scripts/builders/contribute.ts
```diff
@@ -37,10 +37,7 @@ const sourceAssetsSubFolder = "images";
 /* Name of the folder within `outputDirectory` folder where static assets will be copied to */
 const outputAssetsSubFolder = "images";
 /* Base URL to refer to assets from final mdx docs */
-const assetsOutputBaseUrl = path.join(
```
ksen0, line 40: cc @limzykenneth @doradocodes this was breaking the contributor docs

limzykenneth: Commenting it out should be fine. The lint is probably a bit too strict, this should not have been an error lint issue.

